PR #1, opened 2014-11-10
Svav/make python hub

vedstesen:
Adding a new recipe to enabled python on the target. 
The recipe is based on the description from this page: 
http://www.trevorbowen.com/2013/10/07/cross-compiling-python-2-7-5-for-embedded-linux/ 
 And I have used the patch he refers as the basis, but it has been modified slightly. 

I have tested the build on a RaspberryPI running: 
python /usr/lib/python-2.7/test/test___all__.py

### DIFF
--- a/recipes/python/python.inc
+++ b/recipes/python/python.inc
@@ -7,12 +7,12 @@ RECIPE_TYPES = "native machine"
 
 inherit autotools  make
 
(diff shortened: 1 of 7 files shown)


kimrhh:
please submit this against our new meta/python instead :)

kimrhh:
python now lives in meta/python

PR #2, opened 2014-11-24
queue of various old commits

esben:


esben:
These commits are also included in PR #11, so if it is merged, this PR should be closed.

esben:
I removed these commits from PR#11, so this PR is again relevant for merging.

kimrhh:
i don't find the tz stuff clean at all. Patching should be the last resort imo, since it stops us from being able to easily version bump a recipe. We already use DEPENDS:native = "" in many other places for this particular pattern where a native tool is used. We could also just have a seperate recipe for zic (like we have for e.g. qmake).

Another question: is the split useful? seems weird when you then have to download both sources in one of the recipes anyway

esben:
Of-course the split (as in having utilities split from the actual zoneinfo files).  The question is if the recipe should be split or another solution to circular dependency should be implemented.

Does this mean that you are introducing an anti-patch policy for OE-lite/base?

I will pull the tzcode and tzdata from this pr so you have a chance to merge the rest.

kimrhh:
merged to master, thanks!

PR #6, opened 2014-12-17
Gettext: compile without termlibs prefix

diorcety:
Without that option it appends libdir(/usr/lib) in LD_FLAGS so it is conflicting with build machine libraries


kimrhh:
fixed with PR #14

PR #7, opened 2014-12-21
gtk-doc: Add machine recipe type

diorcety:


diorcety:
Hi,

It's needed by some gtk recipes from gtk meta

Regards,

esben:
Is it possible that the recipes needing this are using "gtk-doc" in DEPENDS, where they could just as well use "host-cross:gtk-doc" instead?

diorcety:
Yes maybe that the solution. But actually gtk-doc containing only a m4 (referencing a pkg-config file from somewhere???) That could lead to issues on computer without gtk-doc installed.

esben:
Uh, you are right.  That definitely does sound bad.

kimrhh:
fixed in PR #14

PR #13, opened 2015-04-16
MinGW fixup

esben:


esben:
Could we merge this now?  I would really like to have the master branch include full mingw support so I can use it for regression testing of OE-lite/core.

PR #5, opened 2014-12-16
Libevdev

madsdore:

PR #15, opened 2015-04-17
libevdev: New recipe (rebase of commits from madsdore)

esben:
This PR is a squash and rebase of PR #5.


kimrhh:
is there a reason for adding this to base? I would say it belongs in meta/xorg

esben:
Maybe.  I just rebased the old PR.

For now, I guess it will only be used with X.org.  But won't it also be useful together with libinput and weston/wayland?  And in that case, where would be place recipes used by both X and wayland ?

kimrhh:
without checking, I would think we have other such things i meta/xorg already. I think we should defer a fix for wayland until we get someone willing to add it? Then we could make a meta/freedesktop or something with the shared stuff. IMO we should try to not just add stuff to base because we can't come up with a better place

esben:
Fair enough.

esben:
PR created on xorg.

PR #16, opened 2015-04-27
Fix a couple of kernel 4.x compatibility issues

esben:


esben:
The mingw32 build failure of this PR should be fixed by merging of PR #13.

kimrhh:
so the drawback is that you can't use it for 2.6 kernels anymore if we apply this. Sorry, but I prefer a loud failure that is easy to spot, over a silent not working situation. 

Of course you would build the wrong thing if you tried to build for a 2.6 kernel on a 3.x buildhost.

USE flag with one of two values (2.4 or 2.6+) , and no default? with an error if it is not set

esben:
Building for 2.4 kernels are broken without this patch anyway, unless you build on a 2.4 host.  But I really think we are being highly theoretical here.

esben:
Big fat comment added by the way.

kimrhh:
merged to master, thanks!

PR #17, opened 2015-04-28
Bridge utils recipe, from Debian bridge-utils package

esben:


kimrhh:
almost easier to add this to its own github repo ;) :+1:

PR #22, opened 2015-05-22
Improvements

diorcety:


esben:
The help2man recipe is included in OE-lite/core now, so it should be dropped from this PR.

Villemoes:
This needs to be rebased on current master, if it is still relevant. Otherwise I'm gonna close it.

kimrhh:
i don't mind taking over some of the work in here, would be a shame to drop all the new recipes at least

diorcety:
Hi,
Sorry but currently I have no time to handle this. I may be have more time in the next weeks, but not sure. I'm not using oe-lite frequently now so I am not up to date anymore.
Regards,

PR #24, opened 2015-05-25
Verbose remove

diorcety:


kimrhh:
is this causing problems? I'm not sure why we have it on, but hard to see why that should be a problem

diorcety:
Not a problem at all, but why this recipes an not the others?. a way to trigger a "debu" flag by recipe from the command line and/or a config file should be great

esben:
I am pretty confident that in both recipes, I made the mistake of not removing V=1 before commit.

:+1:

kimrhh:
i'm going to reverse the argumentation here. Actually the default output when running make on this kind of projects, is different from 90% of building software. So it will hide away what is actually being run as commands, just to make it look pretty. I think it makes very little sense to hide away useful information just for the sake of saving a few bytes in a log file. This output is only visible if you bake with -d anyway.

I vastly prefer to minimize the time spent hunting down problems to saving a little bit of disk space. So my solution would be to make V=1 default for this kind of stuff.

PR #35, opened 2015-06-10
Fix and upgrade OpenVPN recipe

esben:


recipes/openvpn/openvpn_2.3.6.oe
```diff
@@ -0,0 +1,9 @@
+# -*- mode:python; -*-
+require ${PN}.inc
+LICENSE = "GPLv2"
+
+EXTRA_OECONF += "--disable-plugin-auth-pam"
+
+EXTRA_OECONF += "--enable-iproute2"
```
kimrhh: am I missing something, or is the iproute2 dep not set up?

kimrhh:
wauv, this is clearly an improvement ;)

PR #36, opened 2015-06-11
intltool: add recipe

diorcety:


kimrhh:
please drop the autoreconf and just use autotools.oeclass

This one is likely not machine type either (correct me if you have a use case for it)

esben:
@diorcety will update this PR with the changes mentioned in @kimrhh's comments?

diorcety:
I will try...

kimrhh:
any updates on this?

diorcety:
Why removing autoreconf? Which isn't better than use provided one?

esben:
There are a couple of downsides to autoreconf:

1. It is slow (as in slower than just autoconf).
2. It is more prone to problems with differences in host tools.

So the general idea is to avoid doing autoreconf, unless special reasons call for it.

diorcety:
Okay i do that with a scope to use minimal host tools. A switch could be a idea to develop.

kimrhh:
could you please explain that a bit more. As a downstream user of some software, the whole idea with autotools is to minimize the need for host tools, by providing a huge configure script you can run to produce all the makefiles. Autoreconf is only needed when you need to touch the input files used to generate that configure script.

I really don't understand how do autoreconf would use less tools from the host?

however, we generally really would like to provide oe-lite recipes for as much as possible. Assuming we had everything needed for running autoreconf, we could do it more safely. I don't see that is currently the case though. Autoreconf is known to break e.g. on RHEL6 since it has really old tools (and we don't stage the needed ones)

But i have feeling we are missing something you are trying to do :) so please explain more, perhaps open a general issue about your idea so we have a better place to track it?

diorcety:
Here my point. This suite of PRs aiming the goal of using the minimal tools from the host. Maybe it was not explicit. I'm sorry. Maybe we have to keep on autoconf and wait for a complete support of that goal to switch on autoreconf

diorcety:
cf https://github.com/diorcety/oe-lite-core/commit/c08d8154932712287ed34313ddbb137dfb30960b

kimrhh:
ok I get that, that is a nice goal and we agree with it very much :)

what I don't get is how that is related to autoconf/autoreconf.

diorcety:
Somewhere in https://github.com/oe-lite/core/pull/13/commits
I'am agree i don't do clear PR ... my bad. Tell me in which order you want i will do 1 commit 1 PR

kimrhh:
i'm only talking about this PR :)

just change the "autotools-autoreconf" to "autotools" and I will merge it. Let us discuss the other ones in that PR instead

diorcety:
done

kimrhh:
merged, thanks! :)

PR #37, opened 2015-06-13
Valgrind upgrade and bugfix

esben:
Upgrade valgrind to 3.10.1 and fix compatibility with glibc 2.21 so we are able to upgrade toolchain.


kimrhh:
could we retest this against master?

esben:
I have merged the toolchain upgrade to OE-lite/core, which should fix this PR.

esben:
There you go, it builds fine with recent gcc and glibc :smiley:

kimrhh:
merged, thanks!

esben:
I always look at the SRC_URI to determine what is being used in a recipe.  Anything else is (like guessing based on files in the recipe dir) is flawed when taking into consideration the way the OE-lite work.  There might be patches left there which are not referred to anymore (these should of-course be removed/cleaned up), and patches could be applied conditionally (not generally a a good idea IMHO, but it is possible and is done in several places).

PR #39, opened 2015-06-19
Zmq

hundeboll:


recipes/libsodium/libsodium.inc
```diff
@@ -0,0 +1,8 @@
+inherit autotools c pkgconfig
```
esben: You should probably inherit "library" class also.  It is designed for single-library recipes such as this.

recipes/libsodium/libsodium.inc
```diff
@@ -0,0 +1,8 @@
...
+
+DESCRIPTION = "libsodium - Sodium is a modern, easy-to-use software library for encryption, decryption, signatures, password hashing and more."
+HOMEPAGE = "https://download.libsodium.org/doc/"
+LICENSE = "ISC"
+DEPENDS_${PN} = "libc"
```
esben: This should be a '+=' assignment, which becomes important when library.oeclass is used, as it has a default DEPENDS_${PN} against ${PN}-dev_${PV}

recipes/libsodium/libsodium.inc
```diff
@@ -0,0 +1,8 @@
...
+RDEPENDS_${PN} = "libc"
```
esben: Should also be '+='

recipes/zeromq/zeromq.inc
```diff
@@ -0,0 +1,18 @@
+inherit autotools pkgconfig c c++ auto-package-utils
+
+DESCRIPTION = "Zeromq - The Intelligent Transport Layer"
+HOMEPAGE = "http://www.zeromq.org"
+LICENSE = "LGPLv3+"
+SRC_URI = "http://download.zeromq.org/zeromq-${PV}.tar.gz"
+
+AUTO_PACKAGE_UTILS = "libzmq curve_keygen"
```
esben: You should remove "libzmq" here, inherit from "auto-package-libs" also, and set AUTO_PACKAGE_LIBS to "zmq", so you get proper library packaging of libzmq.

recipes/zeromq/zeromq.inc
```diff
@@ -0,0 +1,18 @@
...
+PROVIDES_${PN} = "libzmq"
```
esben: Looks strange...

kimrhh:
I have a reworked version of this in one of our projects, I'm going to submit that as soon as I have verified  that it works in the SDK as well

esben:
Sounds good.  I might want to play with 0mq soon, so I would like to avoid duplicating your work.

PR #43, opened 2015-07-03
u-boot-tools: Support FIT image creation

esben:


kimrhh:
do we maybe need to handle this via a flag? thinking for systems that do not have dtc available

esben:
We could do that if you want.
But as we are only doing this for native and sdk, it probably does not matter much.
The dtc binary is less than 80k big, and takes less than 10 seconds to build (all steps) on my laptop.

kimrhh:
oh i see we can always build dtc, fine then

kimrhh:
merged, thanks

PR #103, opened 2015-09-17
Add some USE flags to netperf

esben:


kimrhh:
am I missing something, or is the largefile flag not doing anything? Also, why not hard enable these things if they don't add dependencies?

esben:
Take a look at https://github.com/oe-lite/core/blob/master/classes/largefile.oeclass to see what largefile flag does.

As for hard enabling these features, it might not be the right approach.  As I understand it, enabling them has a negative impact on netperf performance.

kimrhh:
alright, thanks for clearing this up. let's merge it then :)

PR #104, opened 2015-09-17
busybox: Add USE flag to enable '-o loop' support in mount command

esben:

### DIFF
--- a/recipes/busybox/busybox-configure.inc
+++ b/recipes/busybox/busybox-configure.inc
@@ -692,7 +692,7 @@ BUSYBOX_SIMPLE_USE_FLAGS += "mount=1:util/"
 BUSYBOX_SIMPLE_USE_FLAGS += "feature_mount_nfs"
 
 # USE flag: enable feature_mount_loop
-BUSYBOX_SIMPLE_USE_FLAGS += "feature_mount_loop::\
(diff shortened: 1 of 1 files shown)


kimrhh:
looks good, but I'm wondering why we are make all these things so extremely configurable. If this feature where to pull in a huge dependency or something I might understand it, but why not just make this hard enabled by default and then let someone who might need to make a system without it deal with fixing that?

kimrhh:
to be clear, I see nothing wrong with this change (if the option is enabled by default?), but where should we open discussions like this?

esben:
The PR currently does not enable the feature by default.  Should I update with that changed?

As for the general discussion about busybox configurability in OE-lite, I also don't think the current state in OE-lite is ideal. But it clearly (the discussion) belongs somewhere else than in this PR :)

kimrhh:
my personal opinion would be that such a feature should be enabled by default, since that would be what most people would expect from a normal system, so if you don't mind please do that :) I'll start up some issues to discuss some larger things regarding busybox

esben:
I put the enable by default in separate commit, so the first commit can be cherry-picked to old releases without breaking full backwards compatibility.

kimrhh:
oh right, great! thanks

PR #165, opened 2015-09-30
rsync: Do VPATH build

esben:
Hoping that this resolves rebuild issue.


kimrhh:
looks like it did, would still like to know the reason

PR #169, opened 2015-10-06
xloadimage: added recipe of xloadimage

sknsean:
It is currently maintained in debian as patches.


esben:
Perhaps have recipes such as this in OE-lite/X.org instead?  Or perhaps in yet-to-be-created layer?  Or do we want to have various X11 applications here in OE-lite/base?

sknsean:
I'll move it to meta/xorg :-)

PR #173, opened 2015-10-14
Add use flag to set directory to watch for new firmware

ksorensen:
Make the directory to watch for new firmware configurable
via a use flag.

This is necessary for targets where the dupdate archive
is larger than the space available in /tmp.

Signed-off-by: Klaus Henning Sorensen <klaus.sorensen@prevas.dk>


esben:
LGTM.

kimrhh:
merged, thanks

PR #176, opened 2015-10-14
Remove explicit handling of PIDFILE for watchdog app

ksorensen:
In my experience, the busybox watchdog app creates and deletes
the /var/run/watchdog.pid file, and the explicit creation/deletion
of the pidfile in the busybox-watchdog script leads to warning
messages when stopping busybox-watchdog.

### DIFF
--- a/recipes/busybox/files/busybox-watchdog
+++ b/recipes/busybox/files/busybox-watchdog
@@ -1,18 +1,22 @@
 #!/bin/sh
 
+# The busybox watchdog applet creates and removes
+# /var/run/watchdog.pid.
+#
+# It is removed explicitly in stop to be certain
+# it is gone.
+
 PIDFILE=/var/run/watchdog.pid
 WATCHDOG=/dev/watchdog
 
 case "$1" in
     start)
 	splash_progress PLACEHOLDER
-	[ -f $PIDFILE ] \
-	    && kill `cat $PIDFILE`
+	[ -f $PIDFILE ] && kill `cat $PIDFILE`
 	watchdog -t 1 -F $WATCHDOG &
-	echo $! > $PIDFILE
         ;;
     stop)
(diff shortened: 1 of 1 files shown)


esben:
What if busybox-watchdog crashed, then the .pid file would still be there after stop, and the next start would give a warning on the kill command.

Maybe just change the 'rm' to 'rm -f' to avoid the warning?

ksorensen:
/var/run is usually in ram, so normally that use case would not be a problem.

But doing rm -f is fine - it is the warning messages that I found annoying and being extra careful with rm -f is fine and gets rid of the warnings.

ksorensen:
I have force pushed with rm -f in stop and added a comment that the watchdog applet handles creating/deleting the pidfile.

esben:
LGTM :+1:

recipes/busybox/files/busybox-watchdog
```diff
@@ -1,18 +1,22 @@
...
 WATCHDOG=/dev/watchdog
 
 case "$1" in
     start)
 	splash_progress PLACEHOLDER
-	[ -f $PIDFILE ] \
-	    && kill `cat $PIDFILE`
+	[ -f $PIDFILE ] && kill `cat $PIDFILE`
 	watchdog -t 1 -F $WATCHDOG &
-	echo $! > $PIDFILE
```
kimrhh: is this echo really supposed to be removed? If not, please also remove the \ change to avoid later confusion

ksorensen: The echo line is unnecessary, since the busybox watchdog applet creates the pidfile.

The rm is left in to be certain that the pidfile is removed, even though the watchdog applet removes it.

kimrhh:
merged, thanks

PR #178, opened 2015-10-16
busybox: add readlink -f (follow) to readlink use flag

esben:
Cherry-pick of busybox change from PR #167.


kimrhh:
merged, thanks

PR #185, opened 2015-10-19
openssh: Make ssh key generation more flexible and extend with ed25519

esben:
Avoid copy-paste of that 4 line conditional block, and update with current
list of 4 host key types.

### DIFF
--- a/recipes/openssh/files/init
+++ b/recipes/openssh/files/init
@@ -1,6 +1,8 @@
 #! /bin/sh
 set -e
 
+HOST_KEY_TYPES=""
+
 # /etc/init.d/ssh: start and stop the OpenBSD "secure shell" daemon
 
 test -x /usr/sbin/sshd || exit 0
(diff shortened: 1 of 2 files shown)


recipes/openssh/files/init
```diff
@@ -32,18 +34,13 @@ check_config() {
...
-	fi
-	if [ ! -f /etc/ssh/ssh_host_ecdsa_key ]; then
-		echo "  generating ssh ECDSA key..."
-		ssh-keygen -q -f /etc/ssh/ssh_host_ecdsa_key -N '' -t ecdsa
-	fi
+	for keytype in $HOST_KEY_TYPES ; do
+		keyfile="/etc/ssh/ssh_host_"$keytype"_key"
+		if [ ! -f "$keyfile" ]; then
+			echo "  generating ssh $keytype key..."
+			ssh-keygen -q -f "$keyfile" -N '' -t rsa
```
kimrhh: if you add a "sync" here, I can be lazy and not push an additional fixup :) Without, we can lose the key if we poweroff too early (empty file) and then it will not be generated

recipes/openssh/files/init
```diff
@@ -1,6 +1,8 @@
 #! /bin/sh
 set -e
 
+HOST_KEY_TYPES="rsa dsa"
```
kimrhh: why two different defaults? Also, I think we should just drop the insecure dsa key completely. no need to provide bad defaults. Rsa should also only be kept with minimum 4096bit size

recipes/openssh/files/init
esben: Don't you still risk up ending with empty key files?  The race condition has just been made much smaller, but still...
Better solution would be to check if keyfile is zero length, and regenerate it in that case.

recipes/openssh/files/init
esben: Well, the idea is that I don't want to maintain the list in two places, so I decided to leave something different, while still innnocent looking in this location, as it gets rewritten by the USE flag anyway.

recipes/openssh/files/init
kimrhh: no, if you are running e.g. ext4, async is the default. So even when booting just fine all the way through login, after the key was generated, it ends up empty after poweroff. Regenerating it at that point would give the same result. It works where we have seen this problem :)

recipes/openssh/files/init
kimrhh: i think that is going to be confusing. leave it blank, then you would end up with a noisy failure if somehow that list was not populated, instead of silently missing some key types. Or set it to "must-be-configured-by-recipes" :)

kimrhh:
merged, thanks

PR #199, opened 2015-10-27
Fixes for gcc5

sknsean:


recipes/gnupg/gnupg_1.4.19.oe
```diff
@@ -6,8 +6,8 @@ EXTRA_OECONF += "--with-readline"
 
 EXTRA_OECONF += "ac_cv_sys_symbol_underscore=no --without-libusb"
 
-SRC_URI += "file://long-long-thumb.patch"
-SRC_URI += "file://mips_gcc4.4.patch"
+#SRC_URI += "file://long-long-thumb.patch"
+#SRC_URI += "file://mips_gcc4.4.patch"
```
kimrhh: if these are not needed anymore, just remove them :) otherwise forward port is needed. The gcc-4.4 one at least seems not needed anymore since it builds fine

sknsean:
The patches for gnupg classic are applied upstream :-)

recipes/libgpg-error/libgpg-error.inc
```diff
@@ -5,7 +5,28 @@ COMPATIBLE_HOST_ARCHS = ".*linux"
...
+	  arm-*)	      TUPLE=arm-unknown-linux-gnueabi ;;
+	  i586-*|i686-*)  TUPLE=i686-pc-linux-gnu ;;
+	  x86_64-*)     TUPLE=x86_64-pc-linux-gnu ;;
+	  powerpc-*)	TUPLE=powerpc-unknown-linux-gnu ;;
+	  *)          TUPLE=${TARGET_ARCH}-unknown-linux-gnu ;;
+	esac
+
+	cp ${SRCDIR}/${PN}-${PV}/src/syscfg/lock-obj-pub.$TUPLE.h \
+	  ${SRCDIR}/${PN}-${PV}/src/syscfg/lock-obj-pub.$TARGET_FILE.h
+}
```
kimrhh: @esben could you take a look a this? it's stolen from yocto, and seems to fix some of our issues but maybe there is a better way?

esben: Eeew, that was not nice.

For now, I think we should accept the above "solution".  The case probably needs to be extended later on.  And possibly needs to be revisisted when upgrading the recipe :disappointed:

kimrhh:
merged, thanks

PR #200, opened 2015-10-27
watchdog: Added watchdog deamon version 5.14

sknsean:
Recipe and patches is derived from openembedded

Signed-off-by: Sean Nyekjaer <sean@nyekjaer.dk>

### DIFF
--- a/recipes/watchdog/files/fix-ping-failure.patch
+++ b/recipes/watchdog/files/fix-ping-failure.patch
@@ -1,0 +1,79 @@
+Fix ping mode failure
+
+Upstream-Status: Pending  
+
+When watchdog works on ping mode, the system will be rebooted since
+watchdog can not receive the expected ECOREPLY on a setting interval.
+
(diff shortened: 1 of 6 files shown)


recipes/watchdog/files/watchdog-conf.patch
```diff
@@ -0,0 +1,11 @@
...
++++ watchdog-5.13/watchdog.conf	2014-11-13 10:59:43.233943000 +0100
+@@ -20,7 +20,7 @@
+ #test-binary		= 
+ #test-timeout		= 
+ 
+-#watchdog-device	= /dev/watchdog
++watchdog-device	= /dev/watchdog
+ 
+ # Defaults compiled into the binary
+ #temperature-device	=
```
kimrhh: config package instead?

sknsean: good idea

PR #201, opened 2015-10-27
Upstream license fixes

kimrhh:
use correct license specifiers. If the license is not strictly as listed by spdx.org, do not pretend that it is, but instead consistently say "foo-style" to indicate that it is inspired by the foo license


sknsean:
Maybe someone should implement some check ala what they are doing in yocto

kimrhh:
yes please do that :)

PR #205, opened 2015-10-28
ssh: split out conf files, install sysvinit startup files in openssh-sshd pkg

sknsean:


recipes/openssh/openssh.inc
```diff
@@ -70,15 +70,18 @@ do_install_extra_files() {
...
+FILES_${PN}-ssh = "${bindir}/slogin"
 FILES_${PN}-ssh-doc = "${mandir}/*/ssh_config.* ${mandir}/*/slogin.*"
 FILES_${PN}-sshd = "${sysconfdir}/init.d/sshd \
-	${sysconfdir}/ssh/moduli ${sysconfdir}/ssh/sshd_config \
+	${sysconfdir}/rcS.d/* \
+	${sysconfdir}/ssh/moduli \
 	${localstatedir}/run/sshd"
 FILES_${PN}-sshd-doc = "${mandir}/*/moduli.* ${mandir}/*/sshd_config.*"
+FILES_${PN}-ssh-config = "${sysconfdir}/ssh/ssh_config"
+FILES_${PN}-sshd-config = "${sysconfdir}/ssh/sshd_config"
```
kimrhh: you should make the ssh package depend on these two new packages to make sure we don't break stuff downstream. Probably best to let the new packages PROVIDE sshd_config and ssh_config respectively, and then depend on that, so it becomes easy to override those with your own little recipe.

recipes/openssh/openssh.inc
```diff
@@ -72,13 +72,20 @@ inherit auto-package-utils
 AUTO_PACKAGE_UTILS = "ssh-add ssh-agent ssh-keyscan ssh-keygen ssh-keysign \
 	scp ssh sshd sftp sftp-server ssh-pkcs11-helper"
 
+PACKAGES_${PN} += "ssh-config sshd-config"
```
esben: Although not really a requirement, normally we prefix all packages with ${PN}, and I don't see a special reason to do it different here.

sknsean: fixed

esben:
LGTM :+1:

recipes/openssh/openssh.inc
```diff
@@ -72,13 +72,20 @@ inherit auto-package-utils
...
 FILES_${PN}-sshd = "${sysconfdir}/init.d/sshd \
-	${sysconfdir}/ssh/moduli ${sysconfdir}/ssh/sshd_config \
+	${sysconfdir}/rcS.d/* \
+	${sysconfdir}/ssh/moduli \
 	${localstatedir}/run/sshd"
 FILES_${PN}-sshd-doc = "${mandir}/*/moduli.* ${mandir}/*/sshd_config.*"
+FILES_${PN}-ssh-config = "${sysconfdir}/ssh/ssh_config"
+PROVIDES_${PN}-ssh-config = "ssh-config"
+FILES_${PN}-sshd-config = "${sysconfdir}/ssh/sshd_config"
+PROVIDES_${PN}-sshd-config = "sshd-config"
```
sknsean: Is this PROVIDES needed?

esben: No, they are actually both not really needed. You can just provide openssh-ssh-config and/or openssh-sshd-config, which is probably better, as such packages are very unlikely to work with other ssh packages than openssh.

kimrhh:
merged, thanks

PR #208, opened 2015-10-28
busybox update

kimrhh:


recipes/busybox/busybox-1.23.2/defconfig
```diff
@@ -134,7 +134,7 @@ CONFIG_FEATURE_SEAMLESS_GZ=y
 # CONFIG_DPKG_DEB is not set
 # CONFIG_FEATURE_DPKG_DEB_EXTRACT_ONLY is not set
 # CONFIG_GUNZIP is not set
-# CONFIG_GZIP is not set
+CONFIG_GZIP=y
```
esben: You need to drop this.  The gzip command is controlled by USE_busybox_gzip, which is default enabled by the way.  The above change breaks the USE flag.

kimrhh:
replaced by #215

PR #210, opened 2015-10-29
update openssh

kimrhh:


kimrhh:
nope, http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/LICENCE?rev=1.19&content-type=text/x-cvsweb-markup , as you can see they use the term BSD-style in several places as well. I have used "foo-style" consistently for any license that is close to, but not really foo in the licence fixes series #201

esben:
I see that.  But I thought we agreed to follow SPDX for LICENSE specification.

In this case, it seems a bit difficult, with a long list of different licenses.  Not all of them being BSD style btw.

While it won't help being more precise or in compliance with SPDX, we could use "BSD" instead of "BSD-style".  I don't see what meaning the "-style" suffix adds, and even OpenEmbeded (which is kind of dedicated to SPDX) has labeled OpenSSH as LICENSE="BSD".

kimrhh:
we surely agree on using SPDX, but it is not possible when stuff is licensed as something not on the spdx list. In that case I see a number of choices:

1: write something like "BSD" which is not on the list, and also not correct , and quite misleading since it kind of indicates that it is a BSD licence (which it isn't, since then it would be on the list)
2: use a generic way of saying "this licence is similar to something, without being exactly that", that is what "-style" indicates. We could have chosen something else, but the important thing is that this can be used for any licence that is not on the list, but heavily inspired by a license on the list
3: use a package specific license identifier like "openssh", which is more like a lazy way of saying "go look this up yourself" 
4: perhaps indicate that this is not an spdx license with some identifier LICENSE="non-spdx:BSD-4-Clause-style"

I don't see how it matters what OE does, BSD is clearly wrong and misleading in this case, especially since no such license exist.

If there is more than BSD style licenses in the openssh list, then it is an error that it is not added to the LICENSE list, agreed. I would also say it would be better to say "BSD-4-Clause-style" or whatever it best matches instead of just BSD-style

kimrhh:
replaced by #251 , at least for upgrade. License fixup can be in seperate PR

PR #211, opened 2015-10-29
update gnutls and libtasn

kimrhh:


esben:
LGTM, but rebase seems to be needed now.

PR #212, opened 2015-10-31
netbase: Added interface file PROVIDES to allow distro overwrite

sknsean:
Signed-off-by: Sean Nyekjaer <sean@nyekjaer.dk>


esben:
There is no need for this.  Without this change, you can just provide netbase-interfaces from your distro, and arrange for it to be picked up instead of the default netbase-interfaces.  No need to come up with a different name for this.

sknsean:
Okay i See :+1:

PR #214, opened 2015-11-02
Fuse upgrade and packaging fixup

esben:


kimrhh:
heh, ok. Think we should in general try to say something about patch handling in the commits changing them. but that's new policy. Merging, thanks

PR #219, opened 2015-11-14
Mdns

moths:
Install and packaging fixes

### DIFF
--- a/recipes/mdnsd/mdns_544.oe
+++ b/recipes/mdnsd/mdns_544.oe
@@ -1,17 +1,24 @@
 DESCRIPTION = "Bonjour, also known as zero-configuration networking, enables automatic discovery of computers, devices, and services on IP networks."
 HOMEPAGE = "http://developer.apple.com/networking/bonjour/"
 
-#RPROVIDES_${PN} += "libdns_sd.so"
 RECIPE_TYPES = "machine"
 
 COMPATIBLE_HOST_ARCHS = ".*linux"
 
 inherit c make sysvinit
 
+inherit auto-package-libs
+AUTO_PACKAGE_LIBS = "dns_sd nss_mdns"
+
+inherit auto-package-utils
+AUTO_PACKAGE_UTILS = "mdnsd"
+
 SRC_URI = "http://opensource.apple.com/tarballs/mDNSResponder/mDNSResponder-${PV}.tar.gz \
            file://build.patch \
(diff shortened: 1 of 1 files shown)


recipes/mdnsd/mdns_544.oe
```diff
@@ -76,16 +82,9 @@ pkg_prerm_${PN} () {
...
-                    ${includedir}/dns_sd.h "
-
-FILES_${PN}-man += "${mandir}/man8/mdnsd.8 \
-                    ${mandir}/man5/nss_mdns.conf.5 \
-                    ${mandir}/man8/libnss_mdns.8"
-
-PACKAGES = "${PN} ${PN}-dev ${PN}-man ${PN}-dbg"
+FILES_${PN}-libnss-mdns-dev = ""
+FILES_${PN}-dev = ""
+FILES_${PN}-libnss-mdns += "${libdir}/libnss_mdns-0.2.so ${sysconfdir}/nss_mdns.conf"
```
kimrhh: why is it necessary to set these to empty? package order?

recipes/mdnsd/mdns_544.oe
```diff
@@ -76,16 +82,9 @@ pkg_prerm_${PN} () {
...
+FILES_${PN}-libdns-sd-dev = "${includedir}/dns_sd.h"
+FILES_${PN}-libdns-sd += "${libdir}/libdns_sd.so"
```
kimrhh: we usually put the symlink in the -dev package, and the versionen libs in the lib package. any specific reason for not doing the same here?

recipes/mdnsd/mdns_544.oe
moths: Not needed, will push updated without the two empty FILES_ removed

recipes/mdnsd/mdns_544.oe
moths: yes, the .so and .so.1 files do not contain version information as we normaly see with libraries made with e.g. libtool. Linking with -l results in a dependency on the .so file and not the .so.1. Therefore the .so is included in the ${PN}-libdns-sd package. It might be solved later with some manual libtool install(?) on the library or similar.

kimrhh:
could you please leave a comment in the recipe about the .so in -dev ? so we don't risk someone "fixing" it later ;)

PR #222, opened 2015-11-17
gdb: Fixup rdepends, so that "gdb" actually pulls in gdb

esben:


kimrhh:
we really need better ways to deal with the use case where a user just wants the "default" included.

merged, thanks

esben:
Enforcing a default QA failure on empty packages could help a few steps in the right direction...

PR #223, opened 2015-11-20
busybox: udhcpcd.conf in separate package, udhcpd-config

KimBoendergaard:
Will ease making customized BSP's

### DIFF
--- a/recipes/busybox/busybox.inc
+++ b/recipes/busybox/busybox.inc
@@ -73,6 +73,11 @@ do_compile_binfmt_flat_fixup() {
 
 require busybox-install.inc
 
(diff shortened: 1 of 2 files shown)


recipes/busybox/busybox.inc
```diff
@@ -73,6 +73,10 @@ do_compile_binfmt_flat_fixup() {
 
 require busybox-install.inc
 
+PACKAGES:<USE_busybox_udhcpd = "udhcpd-config "
```
kimrhh: looking at the rest of the file, this should be "${PN}-udhcpcd-config ", to be consistent

kimrhh:
merged, thanks

PR #224, opened 2015-12-01
makedevs: apply mode setting on directories

hundeboll:
The mode setting for directories in device table files were ignored in
add_new_directory(). Fix this by adding a call in chmod() in said
function.

### DIFF
--- a/recipes/makedevs/makedevs-1.0.0/makedevs.c
+++ b/recipes/makedevs/makedevs-1.0.0/makedevs.c
@@ -108,6 +108,7 @@ static void add_new_directory(char *name, char *path,
 {
 	mkdir(path, mode);
 	chown(path, uid, gid);
+	chmod(path, mode);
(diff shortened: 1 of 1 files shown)


esben:
Why/how was that ignored?  It (mode) is given to mkdir...

Is it caused by umask?

hundeboll:
from the linux kernel (fs/namei.c):
```
	if (!IS_POSIXACL(path.dentry->d_inode))
		mode &= ~current_umask();
```
So I guess your umask guess is correct...

kimrhh:
@esben ok to merge this then?

esben:
Yes.

In mkdir(2) man page, it reads

> The  argument  mode  specifies  the  mode  for  the  new directory (see
> stat(2)).  It is modified by the process's umask in the usual  way:  ...

Which in this case is probably not what we want, so yes, let's merge this.

PR #226, opened 2015-12-01
kmod: new recipe

hundeboll:


recipes/kmod/kmod_22.oe.sig
```diff
@@ -0,0 +1,2 @@
+42f32b8ae45ad8e3a1da29c9b20af9b5e2e9e676  git://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git;tag=v22
```
esben: I guess this line can be dropped.

kimrhh:
merged, thanks

PR #228, opened 2015-12-15
use upstream openssh configs and update to latest openssh

hundeboll:


recipes/openssh/openssh.inc
```diff
@@ -62,10 +63,15 @@ do_configure_host_key_generation() {
...
 
-do_compile[postfuncs] += "do_compile_install_config_files"
-do_compile_install_config_files () {
-	install -m 0644 ${SRCDIR}/sshd_config ${S}/
-	install -m 0644 ${SRCDIR}/ssh_config ${S}/
+RECIPE_FLAGS += "openssh_kex_extra"
+do_install[postfuncs] += "do_install_config_files"
+do_install_config_files[expand] = 3
+do_install_config_files () {
+	sed -i -E 's!^#(PermitRootLogin) .+!\1 yes!' ${D}${sysconfdir}/ssh/sshd_config
```
kimrhh: this should absolutely not be the default :) it is already easy to override the default config, no reason to make it more insecure

kimrhh:
I have merged PR #205 now. should make your PermitRootLogin and KexAlgorithm changes not needed i think. If not, having them as flags defaulting to the upstream configuration might be ok, just see this as a slippery slope to add handles for all configuration options :)

kimrhh:
replaced by PR #251

PR #231, opened 2015-12-18
Opencv added

moths:


sknsean:
Why not combine this to a single commit?

PR #237, opened 2015-12-21
dhcp: Use upsteam .exmaple conf files and the once provided in the re…

krevsbech:
…cipe.

dhcp: Added conf packages for dhcp-server and dhcp-client

### DIFF
--- a/recipes/dhcp/dhcp4.inc
+++ b/recipes/dhcp/dhcp4.inc
@@ -56,38 +56,45 @@ do_install_extra_files () {
 	install -m 0644 ${SRCDIR}/default-server \
 		${D}${sysconfdir}/default/dhcp-server
 
-	rm -f ${D}${sysconfdir}/dhclient.conf
-	rm -f ${D}${sysconfdir}/dhcpd.conf
-	install -m 0644 ${SRCDIR}/dhclient.conf \
-		${D}${sysconfdir}/dhcp/dhclient.conf
-	install -m 0644 ${SRCDIR}/dhcpd.conf \
-		${D}${sysconfdir}/dhcp/dhcpd.conf
-
+	mv ${D}${sysconfdir}/dhcpd.conf.example \
+	   	${D}${sysconfdir}/dhcp/dhcpd.conf.example
+	mv ${D}${sysconfdir}/dhclient.conf.example \
+	   	${D}${sysconfdir}/dhcp/dhclient.conf.example
 	install -d ${D}${base_sbindir}/
 	mv ${D}${sbindir}/dhclient ${D}${base_sbindir}/
 	install -m 0755 ${S}/client/scripts/linux \
 		${D}${base_sbindir}/dhclient-script
 }
 
-PACKAGES += "${PN}-server ${PN}-client ${PN}-relay ${PN}-omshell"
+PACKAGES += "${PN}-server ${PN}-client ${PN}-relay ${PN}-omshell ${PN}-server-config ${PN}-client-config"
 
 COMMON_DEPENDS = "libdl libcrypto libdl libgcc libssl"
 
 FILES_${PN} = ""
 FILES_${PN}-server = "${sbindir}/dhcpd ${sysconfdir}/init.d/dhcp-server \
-	${sysconfdir}/default/dhcp-server ${sysconfdir}/dhcp/dhcpd.conf"
+	${sysconfdir}/rcS.d/S*dhcp-server"
+# Note that ISC does only provide a .example file, which does not work by default,
+# hence it is kept as an example file
+FILES_${PN}-server-config = "${sysconfdir}/default/dhcp-server \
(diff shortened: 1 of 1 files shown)


kimrhh:
first line too long in commit ad36ff5
seems like the two next commits should be rebased out instead?

kimrhh:
the whole thing needs to be rebased, to not first add stuff then remove it afterwards e.g. bind flags

krevsbech:
Rebased as sugested

KimBoendergaard:
Why put example files in 'config' packages? 
Keeping them in the 'main' package won't conflict with custom config packages like seen in other recipes (where two packages would otherwise provide the same files). But of course they'd take up space at target

kimrhh:
since they are an example of how to generate a proper configuration file, i think it makes a lot of sense. You don't need the example on target, so when you override the config package you only have the working correct config.

kimrhh:
also, we do not want weird empty packages, and we don't want to provide oe-lite.org configuration files, since there is no way of actually providing something that can be used downstream anyway. So when upstream have an example file (and no default conf file) we can provide that instead, and still allow a nice configuration override.

recipes/dhcp/dhcp4.inc
```diff
@@ -53,41 +53,49 @@ do_install_extra_files () {
 		${D}${sysconfdir}/default/dhcp-relay
 	install -m 0755 ${SRCDIR}/init-server \
 		${D}${sysconfdir}/init.d/dhcp-server
-	install -m 0644 ${SRCDIR}/default-server \
+        install -m 0644 ${SRCDIR}/default-server \
```
kimrhh: please drop whitespace changes

recipes/dhcp/dhcp4.inc
```diff
@@ -53,41 +53,49 @@ do_install_extra_files () {
...
 		${D}${sysconfdir}/default/dhcp-server
 
-	rm -f ${D}${sysconfdir}/dhclient.conf
-	rm -f ${D}${sysconfdir}/dhcpd.conf
-	install -m 0644 ${SRCDIR}/dhclient.conf \
-		${D}${sysconfdir}/dhcp/dhclient.conf
-	install -m 0644 ${SRCDIR}/dhcpd.conf \
-		${D}${sysconfdir}/dhcp/dhcpd.conf
+        mv ${D}${sysconfdir}/dhcpd.conf.example \
+    	   	${D}${sysconfdir}/dhcp/dhcpd.conf.example
```
kimrhh: seems like there are so tab/spaces issues here, make sure it is aligned with the current indention

recipes/dhcp/dhcp4.inc
```diff
@@ -53,41 +53,49 @@ do_install_extra_files () {
...
+	mv ${D}${sysconfdir}/dhclient.conf.example \
+		${D}${sysconfdir}/dhcp/dhclient.conf.example
 
-	install -d ${D}${base_sbindir}/
+        install -d ${D}${base_sbindir}/
```
kimrhh: avoid whitespace changes

recipes/dhcp/dhcp4.inc
```diff
@@ -56,38 +56,45 @@ do_install_extra_files () {
 	install -m 0644 ${SRCDIR}/default-server \
 		${D}${sysconfdir}/default/dhcp-server
 
-	rm -f ${D}${sysconfdir}/dhclient.conf
-	rm -f ${D}${sysconfdir}/dhcpd.conf
```
kimrhh: are these files no longer there?

recipes/dhcp/dhcp4.inc
```diff
@@ -56,38 +56,45 @@ do_install_extra_files () {
...
-	install -m 0644 ${SRCDIR}/dhclient.conf \
-		${D}${sysconfdir}/dhcp/dhclient.conf
-	install -m 0644 ${SRCDIR}/dhcpd.conf \
-		${D}${sysconfdir}/dhcp/dhcpd.conf
-
+	mv ${D}${sysconfdir}/dhcpd.conf.example \
+	   	${D}${sysconfdir}/dhcp/dhcpd.conf.example
+	mv ${D}${sysconfdir}/dhclient.conf.example \
+	   	${D}${sysconfdir}/dhcp/dhclient.conf.example
 	install -d ${D}${base_sbindir}/
```
kimrhh: any reason to make these moves? can't we just package them from the original location?

kimrhh:
looks good when we clear up the last small comments/questions

recipes/dhcp/dhcp4.inc
krevsbech: No - They are not put there by upstream

recipes/dhcp/dhcp4.inc
krevsbech: This is done to preserve how the package was placing the files before my changes.

kimrhh:
merged, thanks

PR #238, opened 2015-12-21
bind: Made config package containing the local configuration file nam…

krevsbech:
…ed.conf.local.

### DIFF
--- a/recipes/bind/bind.inc
+++ b/recipes/bind/bind.inc
@@ -9,6 +9,7 @@ inherit autotools sysvinit
 DEPENDS = "libssl libcrypto ${DEPENDS_LIBC}"
 DEPENDS_LIBC = "libdl"
 DEPENDS_LIBC:HOST_LIBC_mingw = ""
+RDEPENDS = "${PN}-config"
 
 SRC_URI = "ftp://ftp.isc.org/isc/bind9/${PV}/${PN}-${PV}.tar.gz"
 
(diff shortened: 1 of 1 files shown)


kimrhh:
commit message style, you need to briefly describe the change in the first line, then add an empty line and then follow up with a longer description.

suggestion for line1: "bind: add config package'

recipes/bind/bind.inc
```diff
@@ -54,6 +55,6 @@ inherit binconfig
 BINCONFIG_FILES = "${bindir}/isc-config.sh"
 FILES_${PN}-dev += "${bindir}/isc-config.sh"
 DEPENDS_${PN}-dev = "libssl libcrypto ${DEPENDS_LIBC}"
-
-PACKAGES =+ "${PN}-utils "
+FILES_${PN}-config = "${sysconfdir}/bind/named.conf.local"
```
kimrhh: are there no other conf files?

krevsbech: Yes there are a number of configs, apart from the .conf.local. However, as I see it we dot he conf package in order to allow later override to provide machine/project specific configuration. And for that purpose only named.conf.local i relevant.
Please consider that including more files also force config packages in project/machine to include the same files.

kimrhh: that's fine, i just need to check that we are not still pushing some oe-lite default configuration package as part of the non-config packages.

krevsbech:
Rewording in git log and added comment in recipe.

kimrhh:
merged, thanks

PR #244, opened 2016-01-10
U boot tools

sknsean:


recipes/u-boot/u-boot-tools.inc
```diff
@@ -13,35 +13,22 @@ S = "${SRCDIR}/u-boot-${PV}"
...
     oe_runmake ${USE_u_boot_tools_config}
 }
 
-MAKE_TARGETS = "tools-all"
-
-do_compile[prefuncs] += "do_compile_fix_gdb_include_path"
-
-do_compile_fix_gdb_include_path() {
-    sed 's#\(BFD_ROOT_DIR =\).*#\1 ${HOST_SYSROOT}${includedir}#' -i ${S}/config.mk
-}
```
kimrhh: are you sure this is no longer needed?

recipes/u-boot/u-boot-tools.inc
```diff
@@ -13,35 +13,22 @@ S = "${SRCDIR}/u-boot-${PV}"
...
-PARALLEL_MAKE = ""
-EXTRA_OEMAKE = 'HOSTCC="${CC}" HOSTSTRIP=":" \
-    HOSTLDFLAGS="${HOST_LDFLAGS}" \
+EXTRA_OEMAKE:machine = ' \
+    LDFLAGS="${LDFLAGS}" \
     CROSS_COMPILE="${HOST_PREFIX}"'
 
 RECIPE_FLAGS = "u_boot_tools_config"
-DEFAULT_USE_u_boot_tools_config = "eNET_config"
+DEFAULT_USE_u_boot_tools_config = "defconfig"
```
kimrhh: that's actually an existing defconfig?

recipes/u-boot/u-boot-tools.inc
```diff
@@ -13,35 +13,22 @@ S = "${SRCDIR}/u-boot-${PV}"
...
+EXTRA_OEMAKE = 'HOSTCC="${CC}" \
+    HOSTLDFLAGS="${HOST_LDFLAGS}"'
 
-DO_COMPILE_FIX_SDK_BUILD = ""
-DO_COMPILE_FIX_SDK_BUILD:sdk = "do_compile_fix_sdk_build"
-do_compile[prefuncs] += "${DO_COMPILE_FIX_SDK_BUILD}"
-
-do_compile_fix_sdk_build() {
-    sed 's#$(obj)./bmp_logo $(LOGO_BMP) >$@#${BUILD_SYSROOT}/bin/bmp_logo $(LOGO_BMP) >$@#' -i ${S}/tools/Makefile
-}
```
kimrhh: tried it in an SDK build?

recipes/u-boot/u-boot-tools.inc
```diff
@@ -13,35 +13,22 @@ S = "${SRCDIR}/u-boot-${PV}"
...
-
-PARALLEL_MAKE = ""
-EXTRA_OEMAKE = 'HOSTCC="${CC}" HOSTSTRIP=":" \
-    HOSTLDFLAGS="${HOST_LDFLAGS}" \
+EXTRA_OEMAKE:machine = ' \
+    LDFLAGS="${LDFLAGS}" \
     CROSS_COMPILE="${HOST_PREFIX}"'
```
kimrhh: sure this is not needed anymore?

recipes/u-boot/u-boot-tools-2015.10/defconfig.patch
```diff
@@ -0,0 +1,11 @@
...
++++ u-boot-2015.10/configs/sandbox_defconfig	2016-01-09 18:12:58.460545297 +0000
+@@ -2,7 +2,7 @@
+ CONFIG_DEFAULT_DEVICE_TREE="sandbox"
+ CONFIG_FIT=y
+ CONFIG_FIT_VERBOSE=y
+-CONFIG_FIT_SIGNATURE=y
++# CONFIG_FIT_SIGNATURE is not set
+ # CONFIG_CMD_IMLS is not set
+ # CONFIG_CMD_FLASH is not set
+ # CONFIG_CMD_SETEXPR is not set
```
kimrhh: where is this used?

recipes/u-boot/u-boot-tools.inc
sknsean: i haven't got a clue what it's doing

recipes/u-boot/u-boot-tools.inc
sknsean: yes, for building tools :-)

recipes/u-boot/u-boot-tools.inc
sknsean: Yes, and it worked just fine

recipes/u-boot/u-boot-tools-2015.10/defconfig.patch
sknsean: The sandbox_defconfig = defconfig :-)
The FIT_SIGNATURE is used for creating signed uboot images for imx6 processors.. When activated the build creates a lot of errors, so maybe the one who needs to sign images can reactivivate the build...

recipes/u-boot/u-boot-tools.inc
kimrhh: then don't just remove it :)

recipes/u-boot/u-boot-tools-2015.10/defconfig.patch
kimrhh: sure, or if you wanted to be a bit more nice to downstream, you could make it possible to drop in their own config :) but, as with many other things in oe-lite, those needed a feature get to do the work

sknsean: Remember these uboot tools are universal, fw_printenv etc... 
Maybe downstream will build their own, customized uboot :-)

recipes/u-boot/u-boot-tools.inc
sknsean: First of all i does not apply ;-)
Again, remember this fix was for a version that is 4 years old... Maybe uboot have fixed their things...
Maybe we should ask @jacobbarsoe what the purpose of this fix was

recipes/u-boot/u-boot-tools-2015.10/defconfig.patch
kimrhh: you made it sound like the signing could be used as part of the generic tools, if this is not the case then there of course is no problem

recipes/u-boot/u-boot-tools.inc
kimrhh: if something no longer applies, you are expected to try to figure out if it is still needed. Otherwise you push all that to me instead :) if the commit introducing it does not explain it, and you can't figure out how to make it work again then we might decide to drop it. But then you add that explanation in your commit message, so it becomes clear to the guy running into the original issue when using your updated recipe, why it was dropped.

In this case git blame reveals "u-boot/u-boot-tools: Fix compile on newer 32-bit distros" so, would probably be nice to test if that is still a problem. If you can't reproduce it, we drop it with a note in the commit msg

recipes/u-boot/u-boot-tools-2015.10/defconfig.patch
sknsean: Yes the signing could indeed be used as a part of the generic tools.
But since no one is using the signing feature yet and it throw errors build time, i think the first person whos gonna work with the signing feature should add a fixup commit :-)

kimrhh: :+1:

recipes/u-boot/u-boot-tools.inc
sknsean: I don't have access to a 32bit machine...
I don't think this patch have anything to do with 32bit.. It fixes gdb include path and maybe we still should apply that (to another file)...
http://git.denx.de/?p=u-boot.git;a=blob_plain;f=tools/gdb/Makefile;hb=HEAD

kimrhh: seems like it should be trivial to make the sed work again, could you do that? It could use some more explanation in a comment though :) looks like something used when running u-boot with gdb

sknsean: I'll try... :-)

kimrhh:
nice work, thanks!

PR #245, opened 2016-01-10
Aarch64 compilefixes

sknsean:


recipes/expat/expat.inc
```diff
@@ -6,6 +6,9 @@ RECIPE_TYPES = "machine native sdk"
 
 inherit autotools make-vpath
 
+#Only used for Aarch64
+AUTOCONFDIR = "/conftools"
```
kimrhh: are you sure it is not picked up in the other builds as well?

recipes/libiconv/libiconv.inc
```diff
@@ -10,6 +10,9 @@ SRC_URI = "ftp://ftp.gnu.org/pub/gnu/libiconv/libiconv-${PV}.tar.gz"
 
 inherit library autotools
 
+#Only used for Aarch64
+AUTOCONFDIR = "/build-aux"
```
kimrhh: are you sure it is not picked up in the other builds as well?

recipes/expat/expat.inc
sknsean: The function of the AUTOCONFDIR can be seen in the aarch64 PR in core :-)

recipes/libiconv/libiconv.inc
sknsean: yes

recipes/expat/expat.inc
kimrhh: ok, but you could just only set that var for that arch, I would much prefer that anyway

sknsean: In core it's only read for aarch64 :-) So i guess i could add :TARGET_ARCH_aarch64, but is allready there in core... We wait for @esbens view of this before i do any more work on this

recipes/libiconv/libiconv.inc
esben: I believe that it is included in the other builds, just not used for anything.

kimrhh:
is this replaced by one of your oe-lite/core PRs ?

sknsean:
No, this is still needed to replace the config.{guess,sub} :-)

recipes/fuse/fuse_2.9.4.oe
```diff
@@ -1,6 +1,7 @@
 require fuse.inc
 
-SRC_URI += "file://dont-run-updaterc.d-on-host.patch"
+SRC_URI += "file://dont-run-updaterc.d-on-host.patch \
+            file://fuse_kernel.h_clean_includes.patch"
```
esben: There is also fuse 2.9.5.  Does that also require this patch?

sknsean: Maybe. It's seems the upstream is in a messy state...
https://github.com/libfuse/libfuse the patch is not applied
http://git.baserock.org/cgit/delta/fuse.git/ the patch is applied

I do not have time to upgrade fuse.

PR #246, opened 2016-01-13
ntp: Added useflag for ntpd options

KimBoendergaard:
I needed my ntpd to be started with -g -x.
Found this to be the best solution

### DIFF
--- a/recipes/ntp/files/ntpd
+++ b/recipes/ntp/files/ntpd
@@ -1,5 +1,8 @@
 #! /bin/sh
 #
+
+ntpd_start_args="NTPD_START_ARGS_PLACEHOLDER"
+
 # ntpd	init.d script for ntpdc from ntp.isc.org
 test -x /usr/bin/ntpd -a -r /etc/ntp.conf || exit 0
 # rcS contains TICKADJ
@@ -33,7 +36,7 @@ stopdaemon(){
 case "$1" in
   start)
   	settick
-	startdaemon -g
+	startdaemon $ntpd_start_args
 	;;
   stop)
   	stopdaemon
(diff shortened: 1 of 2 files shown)


recipes/ntp/files/ntpd
```diff
@@ -41,12 +44,12 @@ case "$1" in
...
   	stopdaemon
   	settick
-	startdaemon -g
+	startdaemon $ntpd_start_args
 	;;
   restart)
   	# Don't reset the tick here
 	stopdaemon
-	startdaemon -g
+	startdaemo $ntpd_start_args
```
esben: Missing an 'n'.

KimBoendergaard:
You are right. Fixed now

recipes/ntp/ntp.inc
```diff
@@ -45,6 +45,14 @@ do_install_extra_files() {
     install -m 755 ${SRCDIR}/ntpd ${D}/${sysconfdir}/init.d/ntpd
 }
 
+RECIPE_FLAGS +="ntpd_start_args"
```
esben: While I do see that there is some amount of redundancy in it, I would like to suggest using the name USE_ntp_ntpd_start_args instead of USE_ntpd_start_args.  The reason for this, is that there are other ntpd implementations out there (fx. the one in busybox), and they are not fully argument compatible, so setting ntpd_start_args in fx. a DISTRO could be problematic.

kimrhh: but when would you ever add two different ntpds to the same rootfs? :)

esben: Most likely never.  But I could very well want to have a single distro that supports two different ntpds.  Actually, I do that now, so I am pretty sure it is a realistic scenario :)

PR #249, opened 2016-01-14
valgrind: new version

sknsean:
Added new version


recipes/valgrind/valgrind-3.10.1/configure-linux-kernel-hardcode.patch
```diff
@@ -1,26 +0,0 @@
...
-+++ valgrind-3.9.0/configure.ac	2015-04-28 08:34:50.472585481 +0200
-@@ -273,7 +273,7 @@
-         # Ok, this is linux. Check the kernel version
-         AC_MSG_CHECKING([for the kernel version])
- 
--        kernel=`uname -r`
-+        kernel=3.0
- 
-         case "${kernel}" in
-              2.6.*|3.*)
```
kimrhh: what happened to this? Is upstream still checking the kernel version on the build host, and we can't specify it? 

if so, we should probably create a warning in the recipe itself, to those that might be building for 2.6 kernels

sknsean:
Yes upstream does indeed still check to kernel version of the build host :-(
But i should not be a problem as they only check if the kernel version is lower than 2.6 :-)
```
        kernel=`uname -r`

        case "${kernel}" in
             0.*|1.*|2.0.*|2.1.*|2.2.*|2.3.*|2.4.*|2.5.*)
                    { $as_echo "$as_me:${as_lineno-$LINENO}: result: unsupported (${kernel})" >&5
$as_echo "unsupported (${kernel})" >&6; }
                    as_fn_error $? "Valgrind needs a Linux kernel >= 2.6" "$LINENO" 5
                    ;;

             *)
                    { $as_echo "$as_me:${as_lineno-$LINENO}: result: 2.6 or later (${kernel})" >&5
$as_echo "2.6 or later (${kernel})" >&6; }
                    ;;
        esac
```

PR #250, opened 2016-01-14
libxml: added python module

sknsean:


recipes/libxml/libxml2.inc
```diff
@@ -11,18 +11,39 @@ DEPENDS_HOST_OS = "libdl"
...
 export LDFLAGS += "${LDFLAGS_LDL}"
 
-PACKAGES =+ "${PN}-utils"
+DO_INSTALL_PYTHONFIXUP = "do_install_pythonfixup"
+DO_INSTALL_PYTHONFIXUP:machine = ""
+DO_INSTALL_PYTHONFIXUP:sdk = ""
+
+#This hack is needed python modules are not allowed to be in the same dir as the module
+#Libtool prevents the python modules being installed elsewhere
+do_install[postfuncs] += "${DO_INSTALL_PYTHONFIXUP}"
```
kimrhh: i don't understand what you are saying here

recipes/libxml/libxml2.inc
```diff
@@ -11,18 +11,39 @@ DEPENDS_HOST_OS = "libdl"
 DEPENDS_HOST_OS:HOST_LIBC_mingw = ""
 DEPENDS_${PN} += "libc libm ${DEPENDS_HOST_OS}"
 
+EXTRA_OECONF_PYTHON = "--with-python-install-dir=${libdir}"
```
kimrhh: hard-enabling python requires python-dev or similar on the build host? where are the native DEPENDS?

recipes/libxml/libxml2.inc
```diff
@@ -11,18 +11,39 @@ DEPENDS_HOST_OS = "libdl"
...
+EXTRA_OECONF_PYTHON:machine = "--without-python"
+EXTRA_OECONF_PYTHON:sdk = "--without-python"
+
+EXTRA_OECONF_DOC = "--with-catalog --with-docbook"
```
kimrhh: DEPENDS ?

recipes/libxml/libxml2.inc
sknsean: python modules are not allowed to stay in the same directory as libpython2.so...

recipes/libxml/libxml2.inc
sknsean: I guess it uses the hosts python2 :-)
And it would be rather difficult to bake using oe-lite without native python2.

recipes/libxml/libxml2.inc
kimrhh:
```
AC_ARG_WITH(python,
[  --with-python[[=DIR]]     build Python bindings if found])
AC_ARG_WITH(python_install_dir,
[  --with-python-install-dir=DIR
                          install Python bindings in DIR])
...
dnl                                                                                                                                                                                                                                                                           
dnl check for python                                                                                                                                                                                                                                                          
dnl                                                                                                                                                                                                                                                                           

PYTHON_VERSION=
PYTHON_INCLUDES=
PYTHON_SITE_PACKAGES=
PYTHON_TESTS=
pythondir=
if test "$with_python" != "no" ; then
...
```

the latter is what you need to take a look at, to figure out what the actual dependencies are. From a brief look it seems to just need the python interpreter, but please verify that, and add it as DEPENDS

seems like there is a configure arg to shove the modules into the right place as well

kimrhh:
much better :)

sorry for not noticing before, but those package names are not really communicating what is in the packages. Why not call them PN-libxml2-python etc? when it's specifically for python.

sknsean:
Updated package names

PR #254, opened 2016-01-15
busybox: Fix print out of bogus error messages when getting DHCP lease

esben:
This re-adds a patch that was dropped when upgrading busybox.
It is still relevant, but probably needs a bit of extra attention to
be acceptable to upstream.

### DIFF
--- a/recipes/busybox/busybox.inc
+++ b/recipes/busybox/busybox.inc
@@ -41,6 +41,7 @@ SRC_URI += "\
 	file://busybox-ifplugd.conf \
 "
 
+SRC_URI:>USE_busybox_udhcpc = " file://udhcp-simple-script-route-del-dev-null.patch"
 PACKAGES:<USE_busybox_udhcpc = "${PN}-udhcpc-scripts "
 
 RECIPE_FLAGS += "busybox_init_shutdown_umount"
--- a/recipes/busybox/files/udhcp-simple-script-route-del-dev-null.patch
+++ b/recipes/busybox/files/udhcp-simple-script-route-del-dev-null.patch
@@ -1,0 +1,13 @@
+Upstream-status: Pending
+
+--- busybox-1.16.1/examples/udhcp/simple.script.orig	2010-06-28 12:49:36.000000000 +0200
++++ busybox-1.16.1/examples/udhcp/simple.script	2010-06-28 12:49:55.000000000 +0200
+@@ -22,7 +22,7 @@
+ 
+ 		if [ -n "$router" ] ; then
+ 			echo "Deleting routers"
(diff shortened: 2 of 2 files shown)


kimrhh:
could we please only remove specific "known to be bogus" error messages instead of risk ending up throwing real errors away? 

Isn't it possible to fix it?

esben:
Probably, that is what I meant by my comment about extra attention.

But in this PR, I am jus re-adding a patch that I believe has been dropped without proper discussion.

sknsean:
Hi i'm just curious i can't find this patch in the older commits... Can you see where it was dropped?

esben:
It was removed together with uprade from 1.21.0 (f65c00437d06aa78d2f36217dc6ff5006b404d57) to 1.22.0 (adf22f5298d310ee0b374dbdbbcbf9c5fa8c24d1), so back in january 2014.

kimrhh:
sorry, not taking back a bad hack, just because it was removed without discussion. Also, you merged the upgrade that dropped it yourself ;) see: https://api.github.com/repos/oe-lite/base (created_at:)

esben:
I don't see how the fact that I merged the upgrade that removed a specific patch should put me in a worse position to get the patch resurrected.  On the contrary, the fact that I acknowledge that the removal was a mistake, if it changes anything, should put me in a special position to get it reverted.

That said, the patch is clearly not the best solution.  But the upstream script is a really bad solution also.

A proper solution would include replacing the route del with something else, probably use of the "ip" command instead, and would most likely result in OE-lite carrying a larger patch than the one we had before.  Unless it is merged upstream of-course...

kimrhh:
it doesn't put you in a worse position, I just don't see it can be an argument for just blindly accepting it back either. 
I think this is a clear case of where downstream feels it is an issue, but are unwilling to spend the effort on fixing it properly, so maybe just apply it downstream? 

My issue with it is that while it might fix something for you (not exactly sure what, just error msg in logs you don't want to see?) then it also very clearly risk silencing real errors for others, and for that reason it seems like a bad idea to have upstream.

You could make it to something that only happens when you set a USE, so it is at least not done by default.

esben:
The issue is only to avoid seeing a false error message in the normal use-case (ie. default gw not set).

And you are right, the patch masks possibly other real error messages.

The whole script is really bad IMHO, and I don't think I will pursue it anymore here.  I will probably write a new custom script instead.

It might be worth noting that the script resides in the examples/ dir of busybox, is not installed by busybox build system, and has not been touched since 2013.  So it does not look like something that upstream has a lot of love for.

kimrhh:
:)

PR #261, opened 2016-01-18
texinfo: add recipe

diorcety:


kimrhh:
think you can fix this by adding native:gettext to DEPENDS. 

I find it highly illogical that that doesn't happen when you "inherit gettext" though :) that class should probably be called nls instead.

maybe we could consider letting our autotools-reconf pull it in. @esben thoughts?

diorcety:
@kimrhh indeed, this modification is provided by https://github.com/diorcety/oe-lite-core/commit/b77d075a2bb761675883d0a0c6e47b77d8bc9481

esben:
@kimrhh I am not sure what you exactly are proposing, but I guess it might very well be a good idea :)  Maybe something for a 6.0.0 PR candidate?

PR #263, opened 2016-01-19
busybox: Added USE flags for customization of ifplugd.conf

KimBoendergaard:
With this update it becomes possible to pre-config ifplugd to e.g. support two networks


esben:
LGTM :+1:

PR #209, opened 2015-10-29
wpa-supplicant: add version 2.4

kimrhh:

PR #265, opened 2016-01-19
wpa-supplicant: add version 2.5

sknsean:


kimrhh:
please base this on PR #209

sknsean:
In the history there seems to be no explanation to why we default are using gnutls?
Default in wpa_supplicant is openssl...
Should we change this?

kimrhh:
probably because only gnutls was available on the target this recipe was originially created for :)

You can make it possible to use either if you prefer openssl, but don't just change to another hardcoded default.

kimrhh:
this PR replaces PR #209

kimrhh:
i'll merge this now, then you can add openssl support in an additional PR :)

PR #268, opened 2016-01-27
libnl: updated to 3.2.27

sknsean:


recipes/libnl/libnl.inc
```diff
@@ -16,7 +16,7 @@ FILES_${PN}-dev += "${libdir}/libnl/cli/*/*.la"
 FILES_${PN}-dbg += "${libdir}/libnl/cli/*/.debug"
 
 inherit auto-package-libs
-AUTO_PACKAGE_LIBS = "nl-3 nl-route-3 nl-nf-3 nl-genl-3 nl-cli-3"
+AUTO_PACKAGE_LIBS = "nl-3 nl-route-3 nl-nf-3 nl-genl-3 nl-cli-3 nl-idiag-3 nl-xfrm-3"
```
esben: Adding these in the .inc file, doesn't seem right, as they are then added for old versions when the new version is cherry-picked, which does not make sense.  And the whole purpose of using an .inc file, and this "Add" + "Remove" commit principle, is to be able to cherry-pick without breaking stuff.

recipes/libnl/libnl.inc
```diff
@@ -28,27 +28,17 @@ nl-class-list nl-cls-add nl-cls-delete nl-cls-list \
 nl-link-list nl-pktloc-lookup nl-qdisc-add \
 nl-qdisc-delete nl-qdisc-list"
 AUTO_PACKAGE_UTILS_DEPENDS += "libc libdl libm libnl-3 libnl-cli-3 libnl-genl-3 \
-			 libnl-nf-3 libnl-route-3 libpthread"
+			 libnl-nf-3 libnl-route-3 libpthread libnl-idiag-3"
```
esben: See my comment on the AUTO_PACKAGE_LIBS change.

recipes/libnl/libnl.inc
```diff
@@ -28,27 +28,17 @@ nl-class-list nl-cls-add nl-cls-delete nl-cls-list \
...
 AUTO_PACKAGE_UTILS_RDEPENDS += "libdl libm libnl-3 libnl-cli-3 libnl-cli-3 \
 			libnl-genl-3 libnl-nf-3 libnl-route-3 \
-			libpthread"
+			libpthread libnl-idiag-3"
```
esben: See my comment on the AUTO_PACKAGE_LIBS change.

recipes/libnl/libnl.inc
```diff
@@ -28,27 +28,17 @@ nl-class-list nl-cls-add nl-cls-delete nl-cls-list \
...
-RDEPENDS_${PN}-libnl-route-3 += "libnl-3"
-DEPENDS_${PN}-libnl-genl-3 += "libnl-3"
-RDEPENDS_${PN}-libnl-genl-3 += "libnl-3"
-DEPENDS_${PN}-libnl-cli-3 += "libnl-genl-3 libnl-nf-3 \
-			libnl-route-3 libnl-3 libdl"
-RDEPENDS_${PN}-libnl-cli-3 += "libnl-genl-3 libnl-nf-3 \
-			libnl-route-3 libnl-3 libdl"
 DEPENDS_${PN} += "libc libm ${DEPENDS_LIBC} libnl-3 libnl-cli-3 \
-		libnl-genl-3 libnl-nf-3 libnl-route-3"
+		libnl-genl-3 libnl-nf-3 libnl-route-3 libnl-idiag-3"
```
esben: See my comment on the AUTO_PACKAGE_LIBS change.

recipes/libnl/libnl.inc
```diff
@@ -28,27 +28,17 @@ nl-class-list nl-cls-add nl-cls-delete nl-cls-list \
...
 RDEPENDS_${PN} += "libm libpthread libc libdl libnl-3 libnl-cli-3 libnl-genl-3 \
-		libnl-nf-3 libnl-route-3"
+		libnl-nf-3 libnl-route-3 libnl-idiag-3"
```
esben: See my comment on the AUTO_PACKAGE_LIBS change.

kimrhh:
i agree with Esben, need to first have that part of the code in the version specific part, then we can move it into .inc when the old version has been removed.

recipes/libnl/libnl_3.2.27.oe
```diff
@@ -2,9 +2,10 @@ require ${PN}.inc
 
 LICENSE = "LGPLv2.1+"
 
+SRC_URI = "https://github.com/thom311/libnl/releases/download/libnl3_2_27/libnl-${PV}.tar.gz"
```
sknsean: We need some kind of feature to parse ${PV} from dots to underscores. Almost all github projects release with these paths

PR #274, opened 2016-02-02
Lz4 recipe

esben:


recipes/lz4/lz4_r131.oe
```diff
@@ -0,0 +1,5 @@
+require ${PN}.inc
+
+SRC_URI += "file://windows-h-include.patch"
+SRC_URI += "file://exeext.patch"
```
kimrhh: shouldn't they all be conditional on mingw?

recipes/lz4/lz4.inc
```diff
@@ -0,0 +1,32 @@
...
+S = "${SRCDIR}/lz4-${PV}"
+
+inherit c make
+
+#EXTRA_OEMAKE = "PREFIX=${prefix} CC='${CC}' DESTDIR=${D} LIBDIR=${libdir} INCLUDEDIR=${includedir}"
+#EXTRA_OEMAKE = "PREFIX=${prefix} CC='${CC}' DESTDIR=${D} LIBDIR=${libdir} INCLUDEDIR=${includedir}"
+#
+#do_install() {
+#	oe_runmake install
+#}
```
kimrhh: remove commented dev code? :)

recipes/lz4/lz4_r131.oe
esben: Good question.

Pro: They are only relevant for mingw.

Con: Developers not doing mingw build themselves will see patch conflicts if we apply mingw (and similar for osx patches) for linux builds also.

Let me know what you want here.

recipes/lz4/lz4.inc
esben: Definitely.

recipes/lz4/lz4_r131.oe
kimrhh: I don't understand your con reasoning :) if you don't build for mingw, and those are applied conditionally, how would that couse a patch conflict? If I added another patch only for Linux builds?

esben: Double (or more) negation sucks :)

Pro applying mingw patches conditionally: They are only relevant for mingw.

Pro applying mingw patches unconditionally: Developers not doing mingw build themselves will see patch conflicts for linux builds also.  Patch is closer to a form that can be accepted by upstream.

esben:
Ready for merge?

PR #285, opened 2016-02-25
mono: add recipe

hundeboll:


kimrhh:
nice! :) but this belongs in a seperate meta/mono layer

kimrhh:
replaced by oe-lite/mono#1

PR #287, opened 2016-02-29
mpdecimal: new recipe

hundeboll:
Needed to build python3

### DIFF
--- a/recipes/mpdecimal/mpdecimal.inc
+++ b/recipes/mpdecimal/mpdecimal.inc
@@ -1,0 +1,29 @@
+DESCRIPTION = "Package for correctly-rounded arbitrary precision decimal floating point arithmetic"
+LICENSE = "BSD-2-Clause"
+HOMEPAGE = "http://www.bytereef.org/mpdecimal/index.html"
+RECIPE_TYPES = "machine native sdk"
+
+inherit autotools library
+
+DEPENDS = "libm"
+SRC_URI = "http://www.bytereef.org/software/${PN}/releases/${PN}-${PV}.tar.gz"
+SRC_URI:>HOST_LIBC_mingw = " file://fix-mingw-lib-names.patch"
+
+LIBRARY_NAME = "libmpdec"
+LIBRARY_VERSION = "2"
+
+# fix FILES for mingw builds
+FILES_${PN}:>HOST_LIBC_mingw = " \
+    ${libdir}/${LIBRARY_NAME}-${PV}.dll \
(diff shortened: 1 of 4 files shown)


recipes/mpdecimal/mpdecimal.inc
```diff
@@ -0,0 +1,23 @@
...
+LICENSE = "BSD-2-Clause"
+HOMEPAGE = "http://www.bytereef.org/mpdecimal/index.html"
+RECIPE_TYPES = "machine native"
+
+inherit c autotools library
+
+DEPENDS = "libm"
+SRC_URI = "http://www.bytereef.org/software/${PN}/releases/${PN}-${PV}.tar.gz"
+
+LIBRARY_NAME = "libmpdecimal"
```
esben: Why not set it to "libmpdec" instead?

recipes/mpdecimal/mpdecimal.inc
```diff
@@ -0,0 +1,23 @@
...
+LIBRARY_VERSION = "2"
+
+FILES_${PN} += "${libdir}/libmpdec.so.${LIBRARY_VERSION} ${libdir}/libmpdec.so.${PV}"
+FILES_${PN}-dev += "${libdir}/libmpdec.so"
+DEPENDS_${PN} = "libc libm"
+RDEPENDS_${PN} = "libc libm"
+PROVIDES_${PN} = "libmpdec libmpdec-${LIBRARY_VERSION}"
+PROVIDES_${PN}[qa] += "allow-missing-soname:libmpdecimal-2"
+PROVIDES_${PN}[qa] += "allow-missing-soname:libmpdecimal"
```
esben: If you change LIBRARY_NAME as suggested above, I think the 3 lines here will not be needed (handled automatically by library.oeclass).

hundeboll: How is it handled when the library uses different names for header vs. library?

E.g.
```
/usr/include/mpdecimal.h
/usr/lib/libmpdec.so.2
/usr/lib/.debug/libmpdec.so.2.4.1
/usr/lib/libmpdec.so.2.4.1
```

hundeboll: Disregard that. I am reworking the name now..

esben: If you haven't figured it out, you can always add files to FILES_${PN} and FILES_${PN}-dev as you like.

esben:
Much better, LGTM :+1:

PR #288, opened 2016-02-29
U boot tools fixup

esben:


recipes/u-boot/u-boot-tools.inc
```diff
@@ -44,9 +44,7 @@ inherit auto-package-utils
 U_BOOT_TOOLS = "bmp_logo mkimage envcrc gen_eth_addr img2srec img2brec.sh \
     jtagconsole netconsole setlocalversion"
 
-GDB_TOOLS:machine = ""
```
kimrhh: aren't you missing a GDB_TOOLS:sdk="" there?

recipes/u-boot/u-boot-tools.inc
```diff
@@ -13,22 +13,22 @@ S = "${SRCDIR}/u-boot-${PV}"
...
+EXTRA_OEMAKE:>machine = " ${EXTRA_OEMAKE_CROSS}"
+EXTRA_OEMAKE:>native = ' HOSTCC="${BUILD_CC}" HOSTLDFLAGS="${BUILD_LDFLAGS}"'
+EXTRA_OEMAKE:>sdk = " ${EXTRA_OEMAKE_CROSS}"
 
 RECIPE_FLAGS = "u_boot_tools_config"
 DEFAULT_USE_u_boot_tools_config = "defconfig"
 
 do_configure () {
-    oe_runmake ${USE_u_boot_tools_config}
+    oe_runmake ${USE_u_boot_tools_config} NO_SDL=1
```
kimrhh: isn't this NO_SDL=1 duplicating the EXTRA_OEMAKE one?

recipes/u-boot/u-boot-tools.inc
esben: Huh?  There might have been missing such a line before this change, but as I am deleting the GDB_TOOLS:machine line, I don't see how/why I should by missing something now?  Could you be more specific?

recipes/u-boot/u-boot-tools.inc
esben: Yes, I will drop this change.

recipes/u-boot/u-boot-tools.inc
mnhu: The install snippet in +58 checks if the tools are built, so no need to do {machine,sdk,native} handling anymore.

PR #289, opened 2016-03-01
wpa-supplicant libressl support

sknsean:


kimrhh:
please describe in your commit changing ${B} why this is needed

PR #290, opened 2016-03-03
Zeromq

esben:
This replaces PR #286


recipes/zeromq/zeromq.inc
```diff
@@ -0,0 +1,21 @@
+inherit autotools pkgconfig c++ auto-package-utils auto-package-libs
+
+RECIPE_TYPES = "machine sdk"
```
esben: Hmm.... Do we really need sdk:zeromq?

kimrhh: it's being used, so yes :)

you see it as a problem?

recipes/zeromq/cppzmq_git.oe
```diff
@@ -0,0 +1,17 @@
...
+
+do_install () {
+    install -d ${D}${includedir}
+    install -m 0755 ${S}/zmq.hpp ${D}${includedir}
+}
+
+PACKAGES = "${PN}"
+FILES_${PN} = "${includedir}"
+DEPENDS_${PN} = "libzmq"
+RDEPENDS_${PN} = "libzmq-dev"
```
kimrhh: this seems rather strange, why did you drop the comment? I strongly prefer to add some information about why we do this

recipes/zeromq/zeromq.inc
esben: No, I just cannot imagine why/how zeromq would make sense in sdk context. May I ask what sdk:zeromq is used for?

recipes/zeromq/cppzmq_git.oe
esben: I removed the comment, because I changed the packaging enough to completely invalidate the comment.

What specifically would you like to have a comment about?  The fact that DEPENDS_${PN} and RDEPENDS_${PN} is different?

recipes/zeromq/zeromq.inc
kimrhh: linking an application built using the sdk to libzmq

recipes/zeromq/cppzmq_git.oe
kimrhh: RDEPENDS on a -dev package is not normal, overriding PACKAGES with just ${PN} is also not default stuff :)

recipes/zeromq/zeromq.inc
esben: Huh, if you build an application using "the sdk", that application is for target/MACHINE, so you would need to link it with machine:libzmq.

kimrhh: you're right, I need more coffee. It might be that we did not know for sure where the curve_gen tool should be executed. I would prefer we live the SDK build since it works and is being used.

recipes/zeromq/cppzmq_git.oe
```diff
@@ -0,0 +1,22 @@
...
+
+# There is just one file, so anything more than just one package is silly
+PACKAGES = "${PN}"
+FILES_${PN} = "${includedir}"
+# For build-time package dependency, we pull in libzmq, which is responsible
+# for pulling in libzmq-dev also
+DEPENDS_${PN} = "libzmq"
+# For run-time package dependency (ie. sdk-image RDEPENDS), we pull in
+# libzmq-dev, which is responsible for pulling in libzmq also
+RDEPENDS_${PN} = "libzmq-dev"
```
kimrhh: aren't we normally leaving this to the SDK recipe? i mean, pulling in the -dev as RDEPENDS

recipes/zeromq/zeromq.inc
esben: Ah, you have a host tool for generating som crypto key.  That makes sense to have in sdk, but could then just as well be needed in native, to build keys directly in OE-lite.

recipes/zeromq/cppzmq_git.oe
esben: Correct.  But this whole recipe is just one single file.  Should we really have two packages for that?

kimrhh: my point is, you should not have libzmq-dev in RDEPENDS there, just libzmq. pulling in -dev for SDK is another thing

kimrhh:
think it is time to try rebasing this against current master, and fix the small details so we can get it ready for 6.0 :)

@esben do you time to finish it or should I take it?

recipes/zeromq/cppzmq_git.oe
esben: Ok,  I will try to make it behave like a normal library.

esben:
I will rebase, fixup, and upgrade to latest versions.

recipes/zeromq/zeromq.inc
```diff
@@ -0,0 +1,30 @@
...
+
+AUTO_PACKAGE_LIBS = "zmq"
+AUTO_PACKAGE_LIBS_DEPENDS += "${COMMON_DEPENDS} ${DEPENDS_LIBSODIUM}"
+AUTO_PACKAGE_LIBS_RDEPENDS += "${COMMON_RDEPENDS} ${DEPENDS_LIBSODIUM}"
+AUTO_PACKAGE_LIBS_DEV_DEPENDS = "${PN}-dev_${PV}"
+AUTO_PACKAGE_LIBS_DEV_RDEPENDS = "${PN}-dev_${PV}"
+LIBRARY_VERSION = "5"
+
+RECIPE_FLAGS += "zeromq_libsodium"
+DEPENDS_LIBSODIUM = "libsodium"
```
kimrhh: isn't this supposed to be empty?

recipes/zeromq/zeromq.inc
```diff
@@ -0,0 +1,30 @@
+inherit autotools pkgconfig c++ auto-package-utils auto-package-libs
+
+RECIPE_TYPES = "machine sdk"
+
+DESCRIPTION = "Zeromq - The Intelligent Transport Layer"
+HOMEPAGE = "http://www.zeromq.org"
+LICENSE = "LGPLv3+"
```
kimrhh: no such spdx.org license :)  "LGPL-3.0+"

kimrhh:
ready to merge after two small fixups :) thanks!

esben:
Two small fixups done.

kimrhh:
merged, thanks

PR #291, opened 2016-03-04
Gnutls 3.4

Villemoes:


Villemoes:
gnutls 3.3.9 no longer builds, I need to make that explicitly depend on nettle 2.*. So ignore this for now.

PR #292, opened 2016-03-04
openssl: upgrade to 1.0.2g

sknsean:


recipes/openssl/openssl_1.0.2g.oe
```diff
@@ -1,8 +1,5 @@
 require openssl.inc
 
-# Parallel make is simply broken for now.
-PARALLEL_MAKE = ""
```
esben: Do you have anything concrete knowledge about parallel build fixes?  Or just re-enabling so we can find out?  Would be nice to know if we should be particularly on the look for such issues...

esben:
:+1:

recipes/openssl/openssl_1.0.2g.oe
sknsean: Arch does not do anything special:
https://projects.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/openssl
Yocto however have a patch:
http://git.yoctoproject.org/cgit.cgi/poky/plain/meta/recipes-connectivity/openssl/openssl/parallel.patch
I works on my builder with -j8 :-)

esben: Ok.  So OpenEmbedded has a patch (taken from Gentoo) for parallel build races. In my book, that clearly wins over your "works on my build with -j8" ;)

esben:
:+1:

recipes/openssl/openssl_1.0.2g.oe
kimrhh: for the record, I will happily accept removing disabling PARALLEL_MAKE if it has been shown to work on the oe-lite buildbot. It simply sucks that if someone finds a (maybe?) parallel make issue using $my-special-j-setting it ends up being disabled for everyone.

that said, when there is a patch, of course we chose that route

PR #293, opened 2016-03-08
Add libcurl dependencies to SDK

johnneren:
When using libcurl the dependencies where not met in the SDK.

I have only verified the SDK with 'USE_curl_ssl option'


recipes/curl/curl.inc
```diff
@@ -24,22 +24,30 @@ EXTRA_OECONF_IPV6:USE_curl_ipv6 = "--enable-ipv6"
 
 RECIPE_FLAGS += "curl_ssl"
 DEPENDS:>USE_curl_ssl = " libssl libcrypto"
+RDEPENDS_${PN}:>USE_curl_ssl = " libssl libcrypto"
+RDEPENDS_${PN}-dev:>USE_curl_ssl = " libssl-dev libcrypto-dev"
```
esben: This might not be necessary after all.  It looks like library.oeclass should add these lib*-dev dependencies based on the RDEPENDS_${PN} lib dependencies.  Could you try removing and verify with oe show that RDEPENDS_curl-dev still contains libssl-dev and libcrypto-dev ?

johnneren: I just tested compilation of curl with a new Linux SDK with all RDEPENDS-dev removed and it works just fine.
I will modify the pull request

esben:
:+1:

kimrhh:
merged, thanks!

guess I should try to ask people to run packageqa with all USE enabled when they add such flags :)

PR #294, opened 2016-03-09
openssh: change home dir and shell for sshd user

Villemoes:
The directory /home/sshd presumably doesn't exist, so set it to
/var/run/sshd which is in any case the working directory for sshd
(it's created by the init script). Also, it doesn't make sense to log
in as sshd, so change its shell to /bin/false.


esben:
:+1:

PR #295, opened 2016-03-09
curl: fix libcurl-dev DEPENDS for libcurl

hundeboll:
When inheriting library, one must be carefull to not overwrite DEPENDS,
as library.oeclass sets up the library dependency for ${PN}-dev.

Fix this in the curl recipe by appending to DEPENDS instead of assigning
to it.


esben:
:+1:

PR #297, opened 2016-03-14
asio: new recipe

hundeboll:
asio is a header only library, so no (R)DEPENDS_${PN} is needed.


recipes/asio/asio.inc
```diff
@@ -0,0 +1,12 @@
+DESCRIPTION = "Asio is a cross-platform C++ library for network and low-level \
+    I/O programming that provides developers with a consistent asynchronous \
+    model using a modern C++ approach."
+HOMEPAGE = "http://think-async.com/Asio"
+LICENSE = "BSL-1.0"
+
+inherit autotools c++
+
+DEPENDS = "boost-dev libpthread openssl"
```
kimrhh: +=

Villemoes:
Isn't it the other way around? I.e., a header-only library doesn't have any build-time dependencies (DEPENDS), since, well, there's nothing being built. But it may very well have package dependencies (DEPENDS_*, RDEPENDS_*), e.g. if one of the headers does a #include <foo.h> there's probably a package dependency on libfoo. Or I'm just totally confused?

kimrhh:
i would say it is correct there can't really be any RDEPENDS involved, but I would also say you are correct it is DEPENDS_${PN} that is really needed here.

Villemoes:
Sure there can. If foo.h declares some symbol available from foo.so, we can certainly end up referencing that symbol and thus also need libfoo at runtime. I'm not saying it's the case here, just speaking generally.

PR #298, opened 2016-03-14
busybox: mdev hotplug loading, dnsd, and example udhcpd.conf

hundeboll:


kimrhh:
looks good to me, just awaiting build check

PR #299, opened 2016-03-14
c-ares: new recipe

hundeboll:
Add a recipe for the c-ares library used for async dns lookups. This is
needed to build nodejs, and thus we add a nodejs patch to make it
compatible.

The patch to make nodejs happy about record_start member of tx-structs
is required for all targets.

Explicitly set CFLAGS to avoid failing configure due to U_FORTIFY_SOURCE
not being specified in CPPFLAGS only.

### DIFF
--- a/recipes/c-ares/c-ares.inc
+++ b/recipes/c-ares/c-ares.inc
@@ -1,0 +1,15 @@
+DESCRIPTION = "c-ares is a C library for asynchronous DNS requests (including name resolves)"
+LICENSE = "MIT"
+RECIPE_TYPES = "machine sdk native"
+
+inherit c autotools library
+
+SRC_URI  = "http://c-ares.haxx.se/download/c-ares-${PV}.tar.gz"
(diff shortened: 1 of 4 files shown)


recipes/c-ares/c-ares.inc
```diff
@@ -0,0 +1,15 @@
...
+inherit c autotools library
+
+SRC_URI  = "http://c-ares.haxx.se/download/c-ares-${PV}.tar.gz"
+SRC_URI += "file://0001-deps-provide-TXT-chunk-info-in-c-ares.patch "
+
+LIBRARY_NAME="libcares"
+LIBRARY_VERSION = "2"
+DEPENDS_${PN} = "libc"
+RDEPENDS_${PN} = "libc"
+
```
kimrhh: += to avoid problems later on

hundeboll:
Closing this, as I am not going to upstream nodejs, so this recipe is not needed.

PR #303, opened 2016-03-14
dmalloc: Add recipe for dmalloc. Version: 5.5.2

hundeboll:
Signed-off-by: Christian Sørensen <christian.braunersorensen@prevas.dk>


recipes/dmalloc/dmalloc_5.5.2.oe
```diff
@@ -0,0 +1,60 @@
+DESCRIPTION = "Debug Malloc Library"
+HOMEPAGE = "http://dmalloc.com"
+LICENSE = "CC-BY-SA-3.0"
+
+SRC_URI = "\
+  http://dmalloc.com/releases/dmalloc-${PV}.tgz \
+  file://configure-pagesize-HACK.patch \
```
kimrhh: ugh, this patch seems like a really bad idea :) could we try figuring out where it is from? same with the other patch

kimrhh:
fails for mingw, COMPATIBLE_HOST_ARCHS is ok with me

recipes/dmalloc/dmalloc_5.5.2.oe
hundeboll: Fixes the same issue, but with a decent description and less intrusive:
http://crosstool-ng.org/hg/crosstool-ng/file/tip/patches/dmalloc/5.5.2/110-block_size.patch

The other patch is needed with libc/compilers that optimizes str(n)dup into a macro that must be undefined before special cased. Maybe I should update the patch with a comment?

kimrhh: would be nice, thanks!

PR #306, opened 2016-03-15
Gnutls 3.4

Villemoes:


kimrhh:
could you try if PARALLEL_MAKE is still broken?

could i pursuade you to update the license identifiers to www.spdx.org identifiers? e.g. "gplv3+" -> "gpl-3.0+"

did you run packageqa after upgrading?

Villemoes:
first two handled by just added commits. yes, packageqa passes.

PR #309, opened 2016-03-15
gyp: new (natiive) recipe

hundeboll:
depends on setuptools, so this is also include in the PR.


recipes/gyp/gyp_git.oe
```diff
@@ -0,0 +1,22 @@
+DESCRIPTION = "GYP can Generate Your Projects."
+LICENSE = "BSD-3"
```
kimrhh: http://spdx.org/licenses/

recipes/setuptools/setuptools.inc
```diff
@@ -0,0 +1,15 @@
+DESCRIPTION = ""
+LICENSE = ""
```
kimrhh: nack, need something in both these fields

PR #312, opened 2016-04-12
Add nobody to /etc/passwd and nogroup to /etc/group

esben:


recipes/base-passwd/base-passwd/passwd
```diff
@@ -1 +1,2 @@
 root:*:0:0:root:/root:/bin/sh
+nobody:*:65534:65534:nobody:/nonexisting:/bin/sh
```
kimrhh: probably better to use /bin/false ?

PR #314, opened 2016-04-19
Khso/GitHub busybox

ksorensen:
Various additions to busybox recipe

### DIFF
--- a/recipes/busybox/files/busybox-syslogd
+++ b/recipes/busybox/files/busybox-syslogd
@@ -1,15 +1,16 @@
-#! /bin/sh
+#!/bin/sh
 
 PIDFILE=/var/run/syslogd.pid
 FILE="-O /var/log/messages"
 #REMOTE="-L -R host[:port]"
 #ROTATE_SIZE="-s SIZE"
 #ROTATE_N="-b N"
+#LOG_LEVEL="-l LEVEL"
 
 case "$1" in
     start)
 	splash_progress PLACEHOLDER
-	syslogd -S $FILE $REMOTE $ROTATE_SIZE $ROTATE_N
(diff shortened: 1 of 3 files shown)


recipes/busybox/busybox.inc
```diff
@@ -43,6 +43,7 @@ SRC_URI += "\
 "
 
 PACKAGES:<USE_busybox_udhcpc = "${PN}-udhcpc-scripts "
+SRC_URI:>USE_busybox_udhcpc = " file://udhcp-simple-script.patch"
```
kimrhh: i don't think this falls under "sane defaults", if output like this is a problem it should be worked around downstream. Another possibility is to make it configurable, and not done by default.

recipes/busybox/files/busybox-syslogd
```diff
@@ -1,17 +1,18 @@
...
-	kill `cat $PIDFILE`
-        ;;
+	start)
+		splash_progress PLACEHOLDER
+		syslogd -S $FILE $REMOTE $ROTATE_SIZE $ROTATE_N $LOG_LEVEL
+		;;
+	stop)
+		kill `cat $PIDFILE`
+		;;
 esac
```
kimrhh: could we drop all the whitespace changes?

kimrhh:
other than the two comments, looks good

recipes/busybox/files/busybox-syslogd
ksorensen: I have removed the white space changes.

recipes/busybox/busybox.inc
ksorensen: This commit is removed.

PR #323, opened 2016-05-11
libxslt: new recipe

kimrhh:


recipes/libxslt/libxslt.inc
```diff
@@ -0,0 +1,20 @@
+DESCRIPTION = "The XSLT C library for GNOME"
+
+RECIPE_TYPES = "native"
+
+inherit autotools pkgconfig auto-package-utils
```
esben: Shouldn't you be using library.oeclass also?

kimrhh: oh right, indeed. I was focused on just getting the tool needed to build the xorg stuff :)

PR #328, opened 2016-06-07
pcre: create binconfig-recipe instead of inheriting binconfig class

hundeboll:
The pcre package creates a pcre-config script used by dependent packages
during their configure step. The this shell script is created in machine
context, it must be made available to dependent packages during their
stage step.

Until now, this was done by letting the pcre recipe inherit the
binconfig class, which adds a call to binconfig_stage_fixup() during the
dependent stage step. However, binconfig_stage_fixup() symlinks into
./stage/cross/bin/, which is not "allowed" in the stage fixup functions.
(They should only touch ./stage.unpack/ and not ./stage ).

Also, the symlink-during-stage fixup approach might fail as
./stage/cross/bin is not guarenteed to be created when
binconfig_stage_fixup() is called.

Fix this mess by creating a whole new recipe (pcre-binconfig) which
installs the symlink in cross context, and add is a dependency to the
libpcre package.

### DIFF
--- a/recipes/pcre/pcre-binconfig.oe
+++ b/recipes/pcre/pcre-binconfig.oe
@@ -1,0 +1,10 @@
+DESCRIPTION = "Recipe to provide a symlink to machine:pcre-config in stage/cross/bin/"
+LICENSE = "MIT"
+
+RECIPE_TYPES = "cross"
+FILES_${PN} = "${bindir}"
+
+do_install() {
+    install -d ${D}${bindir}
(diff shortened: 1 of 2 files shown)


Villemoes:
I fail to understand the problem. This symlinking from cross to machine is obviously very deliberately done, so who says it is not "allowed"? And if indeed it is wrong, why would pcre need special treatment, instead of fixing binconfig.oeclass? You've fixed the 'create dir before symlink' part which certainly needed fixing, but is this PR still relevant?

PR #329, opened 2016-06-07
parted: added recipe for version 3.2

sknsean:
Patch from upstream git


kimrhh:
lgtm! thanks

PR #334, opened 2016-06-20
busybox: stty: Allow baudrates up to 4000000

KimBoendergaard:
Probably due to memory usage the baudrate is coded to make it possible to
store it in a short.
This is done by dividing high baudrates by 256 (shift 8 bits) and raising
MSB to signal the division.
This way baudrates being dividable with 256 and not greater than 8388352
can be stored.

To make support for the 'odd' baudrates of 1500000, 2500000 etc bit 14
(MSB -1) has bin chosen to signal a divisor of 250.

Drawback is that coded max, after being divided by 256, must now not
exceed 0x3FFF causing maximum baudrate to be 4194048
or 4095750 for baudrates dividable by 250

### DIFF
--- a/recipes/busybox/busybox.inc
+++ b/recipes/busybox/busybox.inc
@@ -40,6 +40,7 @@ SRC_URI += "\
 	file://busybox-ifplugd.action \
 	file://busybox-ifplugd.conf \
 	file://busybox-dnsd \
+	file://speed_table.patch \
 "
 
 PACKAGES:<USE_busybox_udhcpc = "${PN}-udhcpc-scripts "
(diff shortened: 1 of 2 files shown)


kimrhh:
I never cease to be amazed at how much hackery is in busybox...

should this really be a version specific patch?

anyway, current version is 1.24.1 :)

KimBoendergaard:
Patch is probably not version specific. Wasn't aware that busybox was upgraded. Have now moved patch to files and then we'll see if it will work with 1.24.1 too

kimrhh:
that build failure is completely unrelated. merging

PR #335, opened 2016-06-21
update boost to more recent version

kimrhh:
also add a boost-full package to make it easy to pull
in everything.


mnhu:
```txt
When you say "also" in a patch changelog, that usually means this should
be split into multiple patches.  I'll take this one as no one cares
about debugfs files, but be careful in the future...
```
 - https://lwn.net/Articles/690431/

PR #336, opened 2016-07-04
logrotate: Added recipe for logrotate 3.9.2.

vmbracvlvm:


recipes/logrotate/logrotate_3.9.2.oe
```diff
@@ -0,0 +1,30 @@
+DESCRIPTION = "The logrotate utility is designed to simplify the administration of log files on a system which generates a lot of log files."
+HOMEPAGE = "https://github.com/logrotate/logrotate"
+LICENSE = "GPL-2.0"
+
+COMPATIBLE_HOST_ARCHS = ".*linux"
+
+SRC_URI = "https://github.com/logrotate/logrotate/archive/${PV}.tar.gz" 
+
+DEPENDS =+ "libpopt"
```
kimrhh: i see a libacl dep in configure.ac, are we just picking that up from the buildhost?

vmbracvlvm: acl support is encapsulated inside a condition that checks for availability of acl. As it isn't available it should default to --with-acl=no. The same goes for selinux.

We could add a flag for setting --with-acl however...but currently the recipe builds a minimal logrotate.

kimrhh: please disable these by default (ECONF), we do not want "automagic" behavior in oe-lite recipes, where changes on a build host can cause a different build :)

at least not when it is easily avoided.

PR #345, opened 2016-08-08
upstream PIL changes: mosquitto

kimrhh:

PR #358, opened 2016-08-19
libwebsockets: add recipe

Villemoes:


kimrhh:
how does this relate to the libwebsockets recipe in #345 ? can we just drop that since this is newer?

kimrhh:
hm, this recipe is not as finished as the other one. please port over the auto-package stuff etc.

Villemoes:
Done.

kimrhh:
LGTM, just need a COMPATIBLE_HOST_ARCHS = ".*linux"

PR #359, opened 2016-08-22
mosquitto: new recipe

Villemoes:
This is a replacement for 6503ca7, obsoleting the last part of PR#345. The old one used a mercurial src_uri, which doesn't really seem suppored in OE-lite after all (raise Exception("hg fetcher not implemented") ....). We also allow configuring uuid and websockets support with use flags.

Passes packageqa with any combination of those use flags.


recipes/mosquitto/mosquitto.inc
```diff
@@ -0,0 +1,62 @@
...
+
+SRC_URI = "https://github.com/eclipse/mosquitto/archive/v${PV}.tar.gz"
+
+S = "${SRCDIR}/mosquitto-${PV}"
+
+inherit c++ make auto-package-libs auto-package-utils
+
+RECIPE_FLAGS = "mosquitto_uuid mosquitto_websockets"
+DEFAULT_USE_mosquitto_uuid = "1"
+DEFAULT_USE_mosquitto_websockets = "0"
```
kimrhh: why default to off? is websockets a "heavy" dep?

recipes/mosquitto/mosquitto_1.4.10.oe
```diff
@@ -0,0 +1,3 @@
+require ${PN}.inc
+LIBRARY_VERSION = "1"
+SRC_URI += "file://0001-dont_strip.patch"
```
kimrhh: what is the reason for this patch?

recipes/mosquitto/mosquitto.inc
Villemoes: Just because that's the out-of-the-box default - the same reason for making uuid default on.

recipes/mosquitto/mosquitto_1.4.10.oe
Villemoes: Exactly what I say in the commit message. If we let the Makefile's install target do the stripping, there are no debug symbols for oe-lite's do_install postfunc do_install_strip to find, making the -dbg packages empty and useless.

kimrhh: right, i don't see the commit msg when viewing all the changes at once. Would you mind adding a patch header here like we have started doing other places:
`#source: where is the patch originating from`
`#reason: why is the patch there`
`#upstream: has it been submitted upstream?`
makes it much easier when we want to upgrade it next time.

recipes/mosquitto/mosquitto.inc
kimrhh: :+1: following upstream defaults makes sense

kimrhh:
thanks

PR #369, opened 2016-08-26
vsftpd: fix set_ftp_home postfunc

Villemoes:
A change in passwd.oeclass means the copy of our passwd file is no
longer where this function expects it to be. Rather than just fixing
the path (which shouldn't contain a hard-coded passwd.d anyway), do
what mysql does and patch it in the SRCDIR before passwd.oeclass gets
its hands on it.


esben:
LGTM

PR #370, opened 2016-08-29
lzma and image format updates

Villemoes:
While working on a few qt recipes (in particular qtwebkit) I had to do a little work on some of our image format recipes, which in turn led me to look at the lzma recipe. That doesn't come with a useful -dev package (a single static library without any accompanying headers). So this replaces the lzma implementation from 7-zip from the one provided by xz-utils, which then also provides us with an xz util.

All new/modified recipes pass packageqa for me.


Villemoes:
I really absolutely hate how github shows the commits out of order. The first commit adds xz, the second removes lzma, and only then comes the web/image related patches.

esben:
It is not out of order, it is just a different order ;) I think it is sorted by commit date.  But I feel the same about it.

kimrhh:
this need to be rebased to not contain what is in #373 now?

Villemoes:
Yes and no. The rest of this PR should just be ignored for now due to the auto*** version issue. I'll get back to it at some point.

PR #371, opened 2016-08-30
bind: fix wrong RDEPENDS variable

hundeboll:
The RDEPENDS variable makes no sense without a trailing packagename (and
it screws up 'oe bake bind -t do_fetchall'. Fix it by adding the correct
'_${PN}'.

### DIFF
--- a/recipes/bind/bind.inc
+++ b/recipes/bind/bind.inc
@@ -9,7 +9,7 @@ inherit autotools sysvinit
 DEPENDS = "libssl libcrypto ${DEPENDS_LIBC}"
 DEPENDS_LIBC = "libdl"
 DEPENDS_LIBC:HOST_LIBC_mingw = ""
-RDEPENDS = "${PN}-config"
+RDEPENDS_${PN} = "${PN}-config"
 
 SRC_URI = "ftp://ftp.isc.org/isc/bind9/${PV}/${PN}-${PV}.tar.gz"
 


KimBoendergaard:
Strange RDEPENDS even affect the build. Would expect it to be ignored.

esben:
There should not be anything using an ${RDEPENDS} variable.  But it is going to impact signature, as we are not explicitly ignoring various "unknown" variables.

esben:
LGTM btw :+1:

kimrhh:
that should probably be noisy in packaga-qa then, if we should try to get rid of it.

but why are we even allowing it if it has no use?

esben:
There are no mechanisms to disallow any variables.  We allow RDEPENDS variable in the same way we are allowing YOU_RANDOM_OTHER_VARIABLE_NAME.

KimBoendergaard:
Of course the variable is valid as any other. It is just scarring it impacts the ability to run do_fetchall (that's how the problem was found).
In other words I think something is using the RDEPENDS variable

kimrhh:
while it might be just another var seen from OE-lite, that does not mean it is not highly confusing since DEPENDS = "foo" is fine. I also have a hard time seeing a hard need for using it as just another var :) hence, we could at least warn when people does it

There is certainly also some issue with fetchall if it is somehow being considered

esben:
RDEPENDS is very much valid for image recipes, where it defines the top-level dependencies for rstage. But I assume the bind recipe is not an image recipe.

esben:
Ah, I found it.  As RDEPENDS has meanings for some (image) recipes, there is the following line in fetch.oeclass:

```
do_fetchall[recdeptask] = "DEPENDS:do_fetchall RDEPENDS:do_fetchall"
```

Which means that a fetchall task recurses both DEPENDS and RDEPENDS.

kimrhh:
ok, so is it really needed to make images special and use RDEPENDS when everywhere else use RDEPENDS_${PN}. My understanding of images is that they are also "just" recipes.

What i'm trying to say, if we could remove a way to make an annoying mistake, that would be really preferable, since there are already too many places where an OE-lite project is fragile.

A not so attractive "solution" would be to only recursive the RDEPENDS stuff for images. but meh!

esben:
RDEPENDS and RDEPENDS_${PN} are two completely different things.

RDEPENDS_${PN} is the list of dependencies that package ${PN} pulls in when added to an rstage.

REPENDS is the list of packages that the recipe requires to be added to rstage.

The difference is thus 100% identical to the difference between DEPENDS and DEPENDS_${PN}.

esben:
The code for handling DEPENDS and RDEPENDS is the same. Both are calling set_stage() in do_stage() and do_rstage() respectively.  Discussions about changing the way RDEPENDS works sounds therefore like a discussion to try and make them different, which will IMHO make them more difficult to understand.

kimrhh:
ok, so when you said above that RDEPENDS is just another variable, it really isn't. 

Would it be a problem to make package-qa issue a warning when you have that in a normal recipe? I don't think people will ever run package-qa on an image recipe

esben:
Well, all variables are just variables.  But yes, it has a special purpose for image recipes.  Sorry about adding more confusion here than necessary.

I guess that would normally not be a problem.  You should add a qa flag to disable the check, for those unfortunate enough to actually do that.

While I completely agrees with the idea of improving quality by making it easier to catch obvious problems, I do feal that this exact issue would not be ranked to high on the list of often made mistakes.

PR #373, opened 2016-08-31
add xz-utils recipe

Villemoes:
xz-utils has replaced/ lzma-utils, and allows a much nicer recipe than the current 7-zip based one (which seems to monkey-patch in some code from the tukaani lzma-utils).


recipes/xz/xz.inc
```diff
@@ -0,0 +1,49 @@
...
+# The COPYING file starts
+#
+#   Different licenses apply to different files in this package. Here
+#   is a rough summary of which licenses apply to which parts of this
+#   package (but check the individual files to be sure!):
+#
+# Boiling that down to a single license tag is not easy. Most files
+# are public domain, the remaining are under some GPL variant.
+
+LICENSE = "GPLv2"
```
kimrhh: GPL-2.0

you will need to add something additional to reflect what is stated above. Users of this recipe will look in bom.txt for what licenses is used

recipes/xz/xz.inc
```diff
@@ -0,0 +1,49 @@
+DESCRIPTION = "XZ Utils is free general-purpose data compression software with a high compression ratio, the successor to LZMA Utils"
```
kimrhh: please break this up into shorter lines

recipes/xz/xz.inc
kimrhh: reading in the upstream repo, you at least need to list LGPL-2.1 and GPL-3.0 and "public domain". The latter is really bad actually, since it means there is no proper licensing and you have no clue if it is ok to use the code.

It seems like the license could even be conditional for the different packages, but this is not generally something we have done

Villemoes: Hm, not sure. LGPL-2.1 is only if the bundled getopt_long implementation gets used, which it won't as long as we're using glibc. The mention of GPL-3 is followed by "None of these files end up in the binaries being built.". That leaves "public domain" for the main binaries, and "GPLv2+" for the *grep, *diff, *less wrappers. I see we have monstrosities like "GPLv2 & (LGPLv2.1 | MPL-1.1 | BSD)" - what does that mean, and how do we properly handle recipes with different licenses for different packages?

PR #374, opened 2016-09-02
dtc: update to latest release

kimrhh:


recipes/dtc/dtc_head.oe
```diff
@@ -1,3 +1,4 @@
 require dtc.inc
 
-SRCREV = ";commit=e4b497f367a3b2ae99cc52089a14a221b13a76ef"
+#release v1.4.1
+SRCREV = ";commit=302fca9f4c283e1994cf0a5a9ce1cf43ca15e6d2"
```
esben: Hvorfor ikke
```
SRCREV = ";tag=v1.4.1"
```
?

kimrhh: because then oe-lite thinks .sig file is needed :) which is probably correct, since you can actually move a tag

esben: Thought so.  But does this then flag a general switch away from using tags to using raw commit ids for OE-lite/base?

kimrhh: no I'll leave that to whoever does the updating work. I don't see a problem with using a tag, this is just easier

PR #376, opened 2016-09-11
iperf: version 3.1.3

sknsean:


recipes/iperf/iperf_3.1.3.oe
```diff
@@ -0,0 +1,3 @@
+require iperf.inc
+
+CFLAGS=""
```
kimrhh: why are you resetting CFLAGS? this is not something we should generally do

sknsean: I can't remember why... I have fixed it, let's see what the builder says without it :smile:

skny-prevas: Now i remember why :-)
aarch64-cortexa53t-linux-gnu-gcc: error: -pg and -fomit-frame-pointer are incompatible

sknsean: Now i remember why :-)
aarch64-cortexa53t-linux-gnu-gcc: error: -pg and -fomit-frame-pointer are incompatible

recipes/iperf/iperf_3.1.3.oe
```diff
@@ -0,0 +1,3 @@
+require iperf.inc
+
+CFLAGS += " -fno-omit-frame-pointer"
```
kimrhh: how about only doing it for Aarch64 if that is the only arch with the issue? Also a comment above with a bit of explanation to help the next poor guy please :)

sknsean: It's an issue on all archs... I'll add something for the next poor guy

sknsean: fixed

PR #378, opened 2016-09-13
busybox: Extend sh history length and enable reverse search

esben:
The reverse search feature was previously enabled, but was (accidentally,
I assume) disabled in the upgrade to 1.24.1.  This commit restores the
previous behavior. A shell without Ctrl-R is just not worth it.

And while at it, let's enlarge the history length to 255 entries instead
of 15 lines.  It makes reverse search much more usable.


skny-prevas:
:+1:

sknsean:
:+1:

PR #379, opened 2016-09-13
mysql: added compatible flag "gcc-older-than-6.x"

sknsean:
GCC 6 series breaks this build.
If someone needs this functionality, they could add fx mariaDB

### DIFF
--- a/recipes/mysql/mysql-tools_5.6.13.oe
+++ b/recipes/mysql/mysql-tools_5.6.13.oe
@@ -11,6 +11,7 @@ SRC_URI += " file://gen_pfs_lex_token.patch "
 
 RECIPE_TYPES = "native"
 COMPATIBLE_HOST_ARCHS = ".*linux"
+COMPATIBLE_IF_FLAGS="gcc_older_than_6.x"
 
 inherit cmake c++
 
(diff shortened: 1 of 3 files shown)


esben:
I think it would be fair to postpone this PR until mariadb recipe is added.

kimrhh:
how about we just make it not build by default, then whoever needs mysql can pick up the work later? just some compatible_if_flag fx

sknsean:
@esben The mariaDB recipe is not easy, I have tried to create it. They use a lot of cmake magic.
I have used 4 hours on my investigation, and i won't invest any more time in this.
If this holds the GCC 6.2 upgrade back fine with me...

recipes/mysql/mysql-tools_5.6.13.oe
```diff
@@ -11,6 +11,7 @@ SRC_URI += " file://gen_pfs_lex_token.patch "
 
 RECIPE_TYPES = "native"
 COMPATIBLE_HOST_ARCHS = ".*linux"
+COMPATIBLE_IF_FLAGS="gcc-older-than-6.x"
```
esben: We usually use underscores and not dashes in USE flag names.

kimrhh:
please describe in the commit msg, or as a comment above the flag why we are adding it. don't think it will be logical for another person looking into why it is not built

PR #383, opened 2016-09-28
busybox: update to version 1.24.2

esben:
Without this, I seem to be unable to decompress files larger than ~1036 bytes!


skny-prevas:
Why not bump the busybox recipe to version 0.25 where this patch is included?

sknsean:
Why not bump the busybox recipe to version 0.25 where this patch is included?

esben:
Busybox 1.25.0 is labelled "unstable", so that is a bit early.
An upgrade to 1.24.2 would be fine though, but I just made this fix locally, so thought it would be better to share it than just sit on it.

recipes/busybox/busybox-1.24.2/default_defconfig
```diff
@@ -0,0 +1,1073 @@
+#
+# Automatically generated make config: don't edit
+# Busybox version: 1.24.1
```
sknsean: Nit picking :-)
Version number does not match the recipe version

esben: Nit picking contest accepted ;)
The comment is correct, it was automatically generated with busybox version 1.24.1.
Updating the comment would therefore be falsifying the evidence.

esben:

PR #384, opened 2016-09-28
can-utils: Fix installation prefix

esben:
The upstream Makefile has PREFIX hard-coded to /usr/local


esben:

PR #385, opened 2016-09-28
busybox: Add USE_busybox_chpst flag

esben:
This enables support to add chpst command back.  It is yet another one
of the fallouts from the upgrade to 1.24.1.

In this case, I chose to keep the default to the new setting, as this is
kind of a special command.

If somebody else prefers chpst to be default enabled, that is also fine with me.
I will just update this PR with that on request.


esben:

PR #389, opened 2016-09-30
coreutils: add version 8.25

sknsean:
This is a small step towards a busybox free embedded system

### DIFF
--- a/recipes/coreutils/coreutils-8.25/amversion.patch
+++ b/recipes/coreutils/coreutils-8.25/amversion.patch
@@ -1,0 +1,18 @@
+--- a/configure.orig	2016-09-30 07:38:16.300328305 +0200
++++ a/configure	2016-09-30 11:12:22.253508183 +0200
+@@ -4544,13 +4544,13 @@
+ 
+ # Some tools Automake needs.
+ 
+-ACLOCAL=${ACLOCAL-"${am_missing_run}aclocal-${am__api_version}"}
++ACLOCAL=${ACLOCAL-"${am_missing_run}aclocal"}
+ 
+ 
+ AUTOCONF=${AUTOCONF-"${am_missing_run}autoconf"}
+ 
+ 
+-AUTOMAKE=${AUTOMAKE-"${am_missing_run}automake-${am__api_version}"}
++AUTOMAKE=${AUTOMAKE-"${am_missing_run}automake"}
+ 
+ 
+ AUTOHEADER=${AUTOHEADER-"${am_missing_run}autoheader"}
--- a/recipes/coreutils/coreutils.inc
+++ b/recipes/coreutils/coreutils.inc
@@ -1,0 +1,51 @@
+DESCRIPTION = "The GNU Core Utilities provide the basic file, shell and text \
+manipulation utilities. These are the core utilities which are expected to exist on \
+every system."
+LICENSE = "GPL-3.0+"
+
+COMPATIBLE_HOST_ARCHS = ".*linux"
+
+RECIPE_TYPES = "native machine"
+
+inherit autotools gettext pkgconfig
+
+require conf/fetch/gnu.conf
+
+SRC_URI = "${GNU_MIRROR}/${PN}/${PN}-${PV}.tar.xz"
+
+RECIPE_FLAGS += "singlebin"
+EXTRA_OECONF:>machine += " --libexecdir=${libdir} --enable-install-program=arch ${EXTRA_OECONF_SINGLEBIN}"
+EXTRA_OECONF_SINGLEBIN = ""
+EXTRA_OECONF_SINGLEBIN:USE_singlebin = "--enable-single-binary"
+
+EXTRA_OECONF:>native += " --without-gmp"
+EXTRA_OECONF:>sdk += " --enable-install-program=arch"
+
+DEPENDS += "gettext libpthread libgmp libcap"
+
+RECIPE_FLAGS += "ssl"
+EXTRA_OECONF += "${EXTRA_OECONF_SSL}"
+EXTRA_OECONF_SSL = "--without-openssl"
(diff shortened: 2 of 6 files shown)


recipes/coreutils/coreutils-8.25/amversion.patch
```diff
@@ -0,0 +1,13 @@
+REASON: OE-Lite buildbot uses debian and they are including automake version 1.11
+
+--- a/configure.orig	2016-09-30 07:38:16.300328305 +0200
++++ a/configure	2016-09-30 07:38:26.287039960 +0200
+@@ -4044,7 +4044,7 @@
+ ac_config_headers="$ac_config_headers lib/config.h:lib/config.hin"
+ 
+ 
+-am__api_version='1.15'
++am__api_version='1.11'
```
esben: What impact does this have for developers using newer automake versions?

recipes/coreutils/coreutils.inc
```diff
@@ -0,0 +1,18 @@
...
+COMPATIBLE_HOST_ARCHS = ".*linux"
+
+inherit autotools gettext pkgconfig
+
+require conf/fetch/gnu.conf
+
+SRC_URI = "${GNU_MIRROR}/${PN}/${PN}-${PV}.tar.xz"
+
+EXTRA_OECONF:>native += " --without-gmp"
+EXTRA_OECONF:>machine += " --enable-install-program=arch --libexecdir=${libdir}"
```
esben: Why not `--libexecdir=${libexecdir}` ? (which is automatically added by autotools class btw.)

recipes/coreutils/coreutils-8.25/amversion.patch
sknsean: It only works with this exact version :-(

recipes/coreutils/coreutils.inc
sknsean: Copy paste from yocto recipe... I'll try to remove it

recipes/coreutils/coreutils.inc
```diff
@@ -0,0 +1,38 @@
...
+
+RECIPE_FLAGS += "ssl"
+EXTRA_OECONF += "${EXTRA_OECONF_SSL}"
+EXTRA_OECONF_SSL = "--without-openssl"
+EXTRA_OECONF_SSL:USE_ssl = "--with-openssl=yes"
+DEPENDS:>USE_ssl = " libssl"
+
+inherit auto-package-utils
+
+AUTO_PACKAGE_UTILS = "[ chown dd expr install mkfifo od readlink sha384sum stty tr uptime \
```
esben: Ouch, I fear we will not handle that '[' command in a good way.

esben: Maybe use `bracket:[` and see if that will work.  That should give you coreutils-bracket instead of coreutils-[ which simply looks too dangerous.

ravi-prevas: Probably even better to put the [ binary together with the test binary.

sknsean: Bracket binary(symlink) is placed together with the test binary

kimrhh:
LGTM, do you want to postpone the "choose what utils to install" work to another PR?

sknsean:
I think its impossible(at least for me) to create an essentials package :-)
If I take the footprint cap on, there is difference to have all utils installed og cherry-picking them as they are symlinks.

sknsean:
In the current state of this recipe will it be possible to RDEPEND on this whole package or do i have to RDEPEND on every util/ in my DISTRO conf?

esben:
Several utils from here are definitely useful additions to busybox based systems. If the problem is deciding on a sensible "esentials" package, why not just split everything into packages, and let RDEPENDS_${PN} depend on all of them?  At least that would make it possible to pick the utilities needed for those that would like that (me included...).  A build-time option for single-binary would be a nice thing also, but could come in a separate PR later on.

recipes/coreutils/coreutils.inc
esben: Good idea.

recipes/coreutils/coreutils.inc
```diff
@@ -0,0 +1,42 @@
...
+		 arch chroot df factor join mknod paste realpath sha512sum sum true users \
+		 base32 cksum dir false kill mktemp pathchk rm shred sync truncate vdir \
+		 base64 comm dircolors fmt link mv pinky rmdir shuf tac tsort wc \
+		 basename dirname fold ln nice pr runcon sleep tail tty who \
+		 cat cp du groups logname nl printenv seq sort tee uname whoami \
+		 chcon csplit echo head ls nohup printf sha1sum split test unexpand yes \
+		 chgrp cut env hostid md5sum nproc ptx sha224sum stat timeout uniq \
+		 chmod date expand id mkdir numfmt pwd sha256sum stdbuf touch unlink"
+
+AUTO_PACKAGE_UTILS_RDEPENDS += "${PN}"
```
esben: What ends up in ${PN} ?  It would be nice if the above could be changed to
```
RDEPENDS_${PN} += "${AUTO_PACKAGE_UTILS_PACKAGES}"
```

so you can just add `coreutils` to a rootfs recipe to get entire coreutils in there.

sknsean: ${PN} contains the coreutils binary and libstdbuf.so. The reset of the packages contains only symlinks

kimrhh:
@esben all your comments addressed now?

esben:
Yes, all my comments have been addressed.

But I guess the RDEPENDS handling could be improved for USE_coreutils_singlebin builds, as you could end up with coreutils-whatever installed with a broken symlink to coreutils binary, as coreutils might not be installed.  Of-course, in that cause you can workaround it by adding coreutils to the image RDEPENDS.

PR #391, opened 2016-09-30
opencv: fix build for gcc 6.x

sknsean:
backported fix from OE


recipes/opencv/opencv/0001-cmake-fix-for-GCC-6.x.patch
```diff
@@ -0,0 +1,48 @@
+Reason: Fixes include c++/stdlib.h issues
+Upstream status: https://github.com/opencv/opencv/pull/7391
```
sknsean: Merged upstream

kimrhh:
👍 for taking the patch upstream

PR #367, opened 2016-08-25
gnupg updates

kimrhh:
replaces PR #317

PR #317, opened 2016-05-01
GnuPG modern 2.1.x added, including update for other gnupg packages

moths:

PR #396, opened 2016-10-15
Gnupg updates

moths:
Replaces #367, that replaced #317


recipes/libgcrypt/libgcrypt.inc
```diff
@@ -23,13 +23,6 @@ EXTRA_OECONF = "--without-pth --disable-asm ${EXTRA_OECONF_CAPABILITIES}"
 EXTRA_OECONF_CAPABILITIES = "--without-capabilities"
 EXTRA_OECONF_CAPABILITIES:USE_libgcrypt_capabilities = "--with-capabilities"
 
-# libgcrypt.pc is added locally and thus installed here
```
kimrhh: I'm a bit skeptical if this can just be dropped without causing breakage, surely someone added this to be able to pick up libgcrypt settings through pkg-config. Is it installed by default now?

kimrhh:
adding this to 7.0.0 milestone since security related fixes are the exception to the freeze rule

recipes/libgcrypt/libgcrypt.inc
moths: Yes, the code removed here is from OE [1], I checked debian, ubuntu and arch and they do not handle the .pc.in file and do therefor not have the libgcrypt.pc file in their dev packages [2-4]. Non of the other gnupg libraries have a .pc file. We might benefit from Not installing/depending on an apparently unused package config file.

[1] http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-support/libgcrypt
[2] https://www.archlinux.org/packages/core/x86_64/libgcrypt/
[3] https://packages.debian.org/stretch/amd64/libgcrypt20-dev/filelist
[4] http://packages.ubuntu.com/xenial/amd64/libgcrypt20-dev/filelist

PR #398, opened 2016-10-16
Updates: libnl, bluez. Fix util-linux, lighttpd. Add busybox_dhclient_ifupdown_cmd_option, util-linux-mount suport

moths:
Updates: libnl, bluez. Fix util-linux, lighttpd. Add busybox_dhclient_ifupdown_cmd_option, util-linux-mount suport.


recipes/libnl/libnl_3.2.28.oe
```diff
@@ -0,0 +1,28 @@
+require ${PN}.inc
+
+LICENSE = "LGPLv2.1+"
```
kimrhh: could you make a "fix license" commit, where this is fixed to use spdx: LGPL-2.1+

recipes/busybox/files/rcS
```diff
@@ -3,6 +3,12 @@
 # This is a minmal rcS file for target startup
 #
 
+# Proc & Sys
+[ ! -d /proc ] && mkdir /proc
+mount -t proc proc /proc
+[ ! -d /sys ] && mkdir /sys
+mount -t sysfs sysfs /sys
```
kimrhh: i see no s6 dep on base-files, so assuming it is doing its own thing

PR #406, opened 2016-10-25
vsftpd upgrade, fix CVE, fix musl build

sknsean:


recipes/vsftpd/vsftpd-3.0.3/gcc6_buildfix.patch
```diff
@@ -0,0 +1,35 @@
+--- vsftpd-3.0.2/ls.c.orig	2016-10-25 10:55:42.156931355 +0000
++++ vsftpd-3.0.2/ls.c	2016-10-25 10:57:45.916206693 +0000
+@@ -245,8 +245,8 @@
+   struct mystr brace_list_str = INIT_MYSTR;
+   struct mystr new_filter_str = INIT_MYSTR;
+   struct mystr normalize_filename_str = INIT_MYSTR;
+-  const char *normname;
```
kimrhh: uhm, where did those fixes come from? removing const is not really a "gcc6 buildfix", it might very well silence an error but that is then likely the wrong fix

skny-prevas: From the patch itself is imposible to see what they are doing...
```
const char *normname;
...
normname = realpath(path, NULL);
...
free(normname);
```
Is i allowed to set a const char * ? GCC6 complaining about freeing the const char *

kimrhh:
:+1:

PR #409, opened 2016-10-26
gnupg small fixes

sknsean:


kimrhh:
verified on glibc build with packageqa

PR #418, opened 2016-10-27
openvpn: add version 2.6.12

sknsean:


recipes/openvpn/openvpn_2.3.12.oe
```diff
@@ -0,0 +1,9 @@
+# -*- mode:python; -*-
+require ${PN}.inc
+LICENSE = "GPLv2"
```
kimrhh: could you fix the license, even though it is not your addition :)

recipes/openvpn/openvpn_2.3.12.oe
```diff
@@ -0,0 +1,9 @@
...
+
+EXTRA_OECONF += "--disable-plugin-auth-pam"
+
+EXTRA_OECONF += "--enable-iproute2"
+export IPROUTE="${base_sbindir}/ip"
```
kimrhh: could you check if this line have any effect at all? doesn't look like it to me

PR #422, opened 2016-10-28
Some image recipe updates

Villemoes:
While working on some qt recipes I was led to update some of our image recipes, as well as add one for libwebp. All three pass packageqa for me.


recipes/libgif/libgif.inc
```diff
@@ -0,0 +1,26 @@
+DESCRIPTION = "GIFLIB is a package of portable tools and library routines for working with GIF images."
+HOMEPAGE = "http://giflib.sourceforge.net/"
+LICENSE = "MIT"
+
+SRC_URI = "http://kent.dl.sourceforge.net/project/giflib/giflib-${PV}.tar.bz2"
+
+inherit autotools library auto-package-utils
+
+S = "${SRCDIR}/giflib-${PV}"
```
kimrhh: isn't that what S is by default?

recipes/libtiff/tiff.inc
```diff
@@ -1,24 +1,61 @@
...
 LICENSE = "BSD-2-Clause*"
-HOMEPAGE = "http://www.remotesensing.org/libtiff/"
 
 RECIPE_TYPES = "machine native"
+SRC_URI = "ftp://download.osgeo.org/libtiff/tiff-${PV}.tar.gz"
 
 COMPATIBLE_HOST_ARCHS = ".*linux"
 
-SRC_URI = "ftp://ftp.remotesensing.org/pub/libtiff/tiff-${PV}.tar.gz"
+inherit autotools-autoreconf c++ auto-package-utils auto-package-libs
```
kimrhh: don't see any need for autoreconf here

recipes/libtiff/tiff.inc
```diff
@@ -1,24 +1,61 @@
...
 
-inherit autotools c++
+RECIPE_FLAGS += "tiff_jpeg tiff_lzma tiff_zlib"
```
kimrhh: i don't see any configure handling of these, are they automagic ? would prefer a --disable/--without when they are disabled if that works with the provided configure

kimrhh:
please rebase this onto current master, i can no longer easily find the triggers in the github webhooks thingy

recipes/libgif/libgif.inc
kimrhh: oh, do not name the recipe libgiff when upstream calls the software giflib, that just makes it harder to find stuff

Villemoes:
All comments should be addressed. Now passing explicit --disable flags in libtiff, appending --enable if USE flag is set. autoreconf dropped, not sure why I added that. libgif -> giflib.

recipes/libwebp/webp.inc
```diff
@@ -0,0 +1,17 @@
+
+SRC_URI = "https://storage.googleapis.com/downloads.webmproject.org/releases/webp/libwebp-${PV}.tar.gz"
+
+inherit auto-package-libs auto-package-utils autotools
+
```
kimrhh: you'll have to disable mingw build in this, using the COMPATIBLE_HOST_ARCH dance

sknsean:
@Villemoes try to add something like this in libtiff
#Only used for Aarch64
AUTOCONFDIRS = "/build/autoconf"

Villemoes:
libtiff now builds on aarch64, at least for me. @kimrhh was this the COMPATIBLE_HOST_ARCHS thing I needed?

kimrhh:
no that was because it also fails for mingw :)

kimrhh:
no that was because it also fails for mingw :)

kimrhh:
no that was because webp fails for mingw :)

PR #424, opened 2016-10-28
gdb: add version 7.11

sknsean:
I know it's one version old, but i think we should wait for some yocto patches before upgrading

### DIFF
--- a/recipes/gdb/files/0001-include-sys-types.h-for-mode_t.patch
+++ b/recipes/gdb/files/0001-include-sys-types.h-for-mode_t.patch
@@ -1,0 +1,30 @@
+From fddd7178915968acf680814411b8b3cb137d0587 Mon Sep 17 00:00:00 2001
+From: Khem Raj <raj.khem@gmail.com>
+Date: Tue, 19 Jan 2016 18:18:52 -0800
+Subject: [PATCH 01/10] include sys/types.h for mode_t
+
+mode_t is used in target.h, so we need to include sys/types.h to get the
+defintion
+
+Signed-off-by: Khem Raj <raj.khem@gmail.com>
+---
+Upstream-Status: Pending
+
+ gdb/gdbserver/target.h | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/gdb/gdbserver/target.h b/gdb/gdbserver/target.h
+index 5af2051..f42c510 100644
+--- a/gdb/gdbserver/target.h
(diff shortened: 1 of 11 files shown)


recipes/gdb/files/0005-Add-support-for-Renesas-SH-sh4-architecture.patch
```diff
@@ -0,0 +1,921 @@
+From 8c5fe58c5a0044ddb517a41b277ed27fb3d3bedc Mon Sep 17 00:00:00 2001
+From: Khem Raj <raj.khem@gmail.com>
+Date: Mon, 2 Mar 2015 02:31:12 +0000
+Subject: [PATCH 05/10] Add support for Renesas SH (sh4) architecture.
```
kimrhh: uhm. why?

recipes/gdb/files/0006-Dont-disable-libreadline.a-when-using-disable-static.patch
```diff
@@ -0,0 +1,50 @@
+From f3932cb2960fd54655a448b13d5a5b80f356f8de Mon Sep 17 00:00:00 2001
+From: Khem Raj <raj.khem@gmail.com>
+Date: Sat, 30 Apr 2016 15:25:03 -0700
+Subject: [PATCH 06/10] Dont disable libreadline.a when using --disable-static
```
kimrhh: is this a problem we actually have or did you just steal all the patches? :)

recipes/gdb/files/0007-use-asm-sgidefs.h.patch
```diff
@@ -0,0 +1,36 @@
+From 048675a915a72989f2613386975730da016e7c5d Mon Sep 17 00:00:00 2001
+From: Andre McCurdy <amccurdy@gmail.com>
+Date: Sat, 30 Apr 2016 15:29:06 -0700
+Subject: [PATCH 07/10] use <asm/sgidefs.h>
+
+Build fix for MIPS with musl libc
```
kimrhh: we have no MIPS, let us please keep patches to a minimum

recipes/gdb/files/0006-Dont-disable-libreadline.a-when-using-disable-static.patch
sknsean: yes from openembedded

kimrhh:
only thing left is to not add the mips patch

PR #428, opened 2016-10-28
wpa_supplicant: add version 2.6

sknsean:


kimrhh:
did you test this with libressl? seems like you do since you added a new patch for it, just checking :)

sknsean:
yes i did. I think the wpa_supp guys added some other openssl specific stuff witch is compatible with libressl

PR #430, opened 2016-11-01
Various fixes to prepare for new build host docker image

kimrhh:


recipes/lftp/lftp.inc
```diff
@@ -36,4 +36,8 @@ EXTRA_OECONF = " \
 --with-gnu-ld \
 --without-libresolv \
 --without-gnutls \
+--disable-ipv6 \
```
esben: Would be nice to respect the USE_ipv6 flag or USE_lftp_ipv6 (yes, we are definitely not very consistent with that...).

esben: Hehe, fair enough.

PR #431, opened 2016-11-01
db: add version 6.2.23

sknsean:

PR #432, opened 2016-11-01
iproute2: add version 4.8.0

sknsean:


recipes/iproute2/iproute2_4.8.0.oe
```diff
@@ -0,0 +1,11 @@
...
+
+SRC_URI="https://www.kernel.org/pub/linux/utils/net/${PN}/${PN}-${PV}.tar.xz"
+
+S="${SRCDIR}/${PN}-${PV}"
+
+DEPENDS += "iptables-dev"
+
+FILES_${PN}-tc += "${libdir}/tc/*"
+FILES_${PN}-dev += "${libdir}/tc/.debug/* \
+                    ${datadir}/bash-completion/completions/tc"
```
kimrhh: bash-completion should not go in -dev package, since they are for tc either add them to the tc package or create a bash-completion

kimrhh:
looking at the "buildsystem" for iproute2 as part of PR #431 , it's clear that we should actually generate the "Config" file ourselves instead of relying on the automatic checks. Would you mind doing that work?

sknsean:
I'll take a look :+1:

sknsean:
This depends on https://github.com/oe-lite/base/pull/400 :smile:

recipes/iproute2/iproute2.inc
```diff
@@ -9,18 +9,25 @@ COMPATIBLE_HOST_ARCHS = ".*linux"
...
-SRC_URI = "git://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git;commit=${GIT_COMMIT}"
-S = "${SRCDIR}/iproute2"
-
-DEPENDS += "libdl libresolv libm libdb4 libnl-3 native:bison native:flex"
-
-do_compile[prefuncs] += "do_compile_add_libnl"
-do_compile_add_libnl () {
-	sed -i -e 's#^\(CFLAGS.*\)#\1 -I\${STAGE_DIR}\/\${HOST_TYPE}\/usr\/include\/libnl3\/#' Makefile
+SRC_URI="https://www.kernel.org/pub/linux/utils/net/${PN}/${PN}-${PV}.tar.xz"
+S="${SRCDIR}/${PN}-${PV}"
```
kimrhh: isn't this what S is by default?

PR #433, opened 2016-11-02
libell: add recipe

sknsean:
This builds master as there is no release's yet


recipes/libell/libell_git.oe
```diff
@@ -0,0 +1,5 @@
+require libell.inc
+
+COMMIT = "branch=master"
```
kimrhh: please use a commit so we don't end up with random build failures with this recipe

PR #445, opened 2016-11-03
dhcp: add version 4.3.5 (spring cleaning)

sknsean:


recipes/dhcp/dhcp_4.3.5.oe
```diff
@@ -0,0 +1,9 @@
+require dhcp4.inc
+
+SRC_URI += "file://dhclient-script-drop-resolv.conf.dhclient.patch \
+            file://replace-ifconfig-route.patch \
+"
+SRC_URI += "${SRC_URI_EXTRA}"
+
+DEPENDS += "bind-dev"
+EXTRA_OECONF += "--with-libbind=${TARGET_SYSROOT}${target_libdir}/../ --with-randomdev=/dev/random"
```
kimrhh: that /../ is very suspicious. why is that needed. smells like something is being packaged wrong in bind

sknsean: make[1]:` *** No rule to make target '/build/tmp/work/machine/aarch64-cortexa53t-linux-gnu/dhcp-4.3.5/stage/machine/usr/lib/lib/libirs.a', needed by 'svtest'.  Stop.

PR #174, opened 2015-10-14
Add busybox use flags to include chroot command

ksorensen:
Signed-off-by: Klaus Henning Sorensen <klaus.sorensen@prevas.dk>

PR #447, opened 2016-11-03
gmp.oe: add version 6.1.1 recipe

vmbracvlvm:
From https://github.com/oe-lite/core/pull/174

This version (among other things) adds support for thumb-less arm architecture.


kimrhh:
i don't really understand, why do you want to add this in both meta/core and meta/base?

PR #451, opened 2016-11-10
sqlite updates

Villemoes:


recipes/sqlite/sqlite3.inc
```diff
@@ -25,18 +37,21 @@ EXTRA_OECONF_READLINE = "--disable-readline"
 EXTRA_OECONF_READLINE:USE_sqlite3_readline = "--enable-readline"
 EXTRA_OECONF += "${EXTRA_OECONF_READLINE}"
 DEPENDS:>USE_sqlite3_readline = " libreadline"
+DEPENDS_sqlite3-sqlite3:>USE_sqlite3_readline = " libreadline"
```
kimrhh: think it is a bit more "clean" and like we do elsewhere, to add these to the AUTO_PACKAGE_*_{R}DEPENDS vars instead

recipes/sqlite/sqlite3.inc
```diff
@@ -25,18 +37,21 @@ EXTRA_OECONF_READLINE = "--disable-readline"
...
 
 RECIPE_FLAGS += "sqlite3_extensions"
 EXTRA_OECONF_EXTENSIONS = "--disable-dynamic-extensions"
 EXTRA_OECONF_EXTENSIONS:USE_sqlite3_extensions = "--enable-dynamic-extensions"
 EXTRA_OECONF += "${EXTRA_OECONF_EXTENSIONS}"
 DEPENDS:>USE_sqlite3_extensions = " libdl"
+DEPENDS_sqlite3-sqlite3:>USE_sqlite3_extensions = " libdl"
+RDEPENDS_sqlite3-sqlite3:>USE_sqlite3_extensions = " libdl"
+DEPENDS_sqlite3-libsqlite3:>USE_sqlite3_extensions = " libdl"
+RDEPENDS_sqlite3-libsqlite3:>USE_sqlite3_extensions = " libdl"
```
kimrhh: same as above

PR #456, opened 2016-11-11
bind: USE flag control of libz dependency

esben:
Without this, I see:

    checking for json library... no
    checking for zlib library... yes
    checking for library containing deflate... no
    configure: error: found zlib include but not library.
    Error: Command failed: 'LC_ALL=C ..../bind-9.11.0-P1/tmp/do_configure.20161111151724.run': 1

### DIFF
--- a/recipes/bind/bind.inc
+++ b/recipes/bind/bind.inc
@@ -28,6 +28,16 @@ EXTRA_OECONF = "\
     --with-export-libdir=${libdir} \
 "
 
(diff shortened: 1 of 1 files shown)


sknsean:
Our upstream builder see this:
`checking for json library... no
checking for zlib library... no
`
Could it be configure that finds a native zlib.h on your builder and tries to link with it?

esben:
@sknsean Sounds likely. And yes, better to create a USE flag and control zlib enable/disable that way.

PR #458, opened 2016-11-16
ofono and mobile-broadband-provider-info

Villemoes:


sknsean:
@lkmikkel do you have any comments to this ?

recipes/ofono/ofono.inc
```diff
@@ -0,0 +1,58 @@
...
+RECIPE_FLAGS += "ofono_test"
+EXTRA_OECONF += "--disable-test"
+EXTRA_OECONF:>USE_ofono_test = " --enable-test"
+DEFAULT_USE_ofono_test = False
+
+RECIPE_FLAGS += "ofono_tools"
+EXTRA_OECONF += "--disable-tools"
+EXTRA_OECONF:>USE_ofono_tools = " --enable-tools"
+DEFAULT_USE_ofono_tools = False
+
```
lkmikkel: It would be nice if was possible to disable provisioning (mobile-broadband-provider-info).

Maybe like this

RECIPE_FLAGS += "ofono_provision"
EXTRA_OECONF += "${EXTRA_OECONF_PROV}"
EXTRA_OECONF_PROV = "--disable-provision"
EXTRA_OECONF_PROV:USE_ofono_provision = "--with-provisiondb=${datadir}/mobile-broadband-provider-info/serviceproviders.xml"
DEPENDS:>USE_ofono_provision = " mobile-broadband-provider-info"
DEPENDS_${PN}:>USE_ofono_provision = " mobile-broadband-provider-info"
RDEPENDS_${PN}:>USE_ofono_provision = " mobile-broadband-provider-info"
DEFAULT_USE_ofono_provision = True

kimrhh: is that really worth the "pain"? I mean, my understanding is we are talking about one not-too-large xml file :)

recipes/mobile-broadband-provider-info/mobile-broadband-provider-info.inc
```diff
@@ -0,0 +1,6 @@
+DESCRIPTION = "mobile broadband settings for different service providers in different countries"
+LICENSE = "CC-PD"
+
+SRC_URI = "https://git.gnome.org/browse/mobile-broadband-provider-info/snapshot/${PN}-${PV}.tar.xz"
+
+inherit autotools-autoreconf
```
kimrhh: reconf?

recipes/ofono/ofono.inc
```diff
@@ -0,0 +1,58 @@
+DESCRIPTTION = "oFono is a free software project for mobile telephony (GSM/UMTS) applications"
+LICENSE = "GPLv2"
```
kimrhh: spdx

recipes/ofono/ofono.inc
lkmikkel: You are right Kim, is not worth the pain. Disabling the plugin can be done in the initscript.

recipes/mobile-broadband-provider-info/mobile-broadband-provider-info.inc
Villemoes: Yep, the tarball doesn't come with a configure script, so we need to run autoreconf.

PR #463, opened 2016-11-25
add php recipe

Villemoes:


recipes/php/php.inc
```diff
@@ -0,0 +1,85 @@
...
+DEPENDS_SSL:USE_php_ssl = "libssl libcrypto"
+DEPENDS += "${DEPENDS_SSL}"
+DEPENDS_${PN} += "${DEPENDS_SSL}"
+RDEPENDS_${PN} += "${DEPENDS_SSL}"
+
+RECIPE_FLAGS += "php_sockets"
+DEFAULT_USE_php_sockets = True
+EXTRA_OECONF_SOCKETS = "--disable-sockets"
+EXTRA_OECONF_SOCKETS:USE_php_sockets = "--enable-sockets"
+EXTRA_OECONF += "${EXTRA_OECONF_SOCKETS}"
```
kimrhh: any reason to make this optional when it does not pull in additional deps?

recipes/php/php.inc
```diff
@@ -0,0 +1,85 @@
...
+
+RECIPE_FLAGS += "php_session"
+DEFAULT_USE_php_session = True
+EXTRA_OECONF_SESSION = "--disable-session"
+EXTRA_OECONF_SESSION:USE_php_session = "--enable-session"
+EXTRA_OECONF += "${EXTRA_OECONF_SESSION}"
```
kimrhh: same as above

Villemoes: Probably not, but there may be some who want to just use php as an interpreter on the target without using it with a webserver, so in that case it might make sense to disable the webby things. I actually almost had this is a use case, but that was then dropped for other reasons.

PR #466, opened 2016-12-01
libpcap: enable native build

Villemoes:


Villemoes:
Why? In any case the first patch obviously belongs here.

kimrhh:
because (s)RTP is a streaming protocol

PR #474, opened 2016-12-20
boost.inc: remove stray do_fetch prefunc

Villemoes:
First of all, the addhook statement has the side-effect of setting the
emit flag to "" for the function in question, so add_deps is not
available when do_fetch runs, so this does nothing. (When asked to run a
non-existing function, OE-lite substitutes an always-successful noop
function.) I also fail to see what setting an RDEPENDS_ variable has to
do with do_fetch.


kimrhh:
indeed seems a bit weird, probably a leftover from trying to make the function run early enough or just some copy-pasta failure. thanks

PR #476, opened 2016-12-22
bzip2: use AUTO_PACKAGE_UTILS_SYMLINKS

Villemoes:
Using AUTO_PACKAGE_UTILS_SYMLINKS_* automatically does these FILES_* and
PROVIDES_* settings.

### DIFF
--- a/recipes/bzip2/bzip2.inc
+++ b/recipes/bzip2/bzip2.inc
@@ -18,12 +18,8 @@ AUTO_PACKAGE_UTILS_SYMLINKS_bzdiff = "bzcmp"
 AUTO_PACKAGE_UTILS_SYMLINKS_bzgrep = "bzegrep bzfgrep"
 AUTO_PACKAGE_UTILS_SYMLINKS_bzmore = "bzless"
 
-RDEPENDS_${PN} = "${PN}-bzip2 ${PN}-bzdiff ${PN}-bzgrep ${PN}-bzip2recover ${PN}-bzmore"
-RDEPENDS_${PN}-bzip2 = "${PN}-libbz2"
-RDEPENDS_${PN}-bzdiff = "${PN}-libbz2"
-RDEPENDS_${PN}-bzgrep = "${PN}-libbz2"
-RDEPENDS_${PN}-bzip2recover = "${PN}-libbz2"
-RDEPENDS_${PN}-bzmore = "${PN}-libbz2"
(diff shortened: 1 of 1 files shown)


recipes/bzip2/bzip2.inc
```diff
@@ -13,27 +13,13 @@ FILES_${PN}-libbz2-dev = "${includedir} ${libdir}/pkgconfig/bzip2.pc"
...
-RDEPENDS_${PN}-bzdiff = "${PN}-libbz2"
-RDEPENDS_${PN}-bzgrep = "${PN}-libbz2"
-RDEPENDS_${PN}-bzip2recover = "${PN}-libbz2"
-RDEPENDS_${PN}-bzmore = "${PN}-libbz2"
+AUTO_PACKAGE_UTILS_SYMLINKS_bzip2 = "bunzip2 bzcat"
+AUTO_PACKAGE_UTILS_SYMLINKS_bzdiff = "bzcmp"
+AUTO_PACKAGE_UTILS_SYMLINKS_bzgrep = "bzegrep bzfgrep"
+AUTO_PACKAGE_UTILS_SYMLINKS_bzmore = "bzless"
+
+AUTO_PACKAGE_UTILS_RDEPENDS = "libbz2"
```
esben: While this is theoretically better/correct, I always fear that a utility like bzip2 could turn out to depend on some unofficial library ABI, making it incompatible with alternative libraries.

Villemoes: I think I see what you mean, but I don't think libbz2 (the canonical one) has any kind of official API that other implementers could or would try to mimick. And that actually goes for most of the libraries-that-also-come-with-a-few-utils that we build. E.g. sqlite3 and pcregrep both depend on the library provided by their recipe. So I can definitely keep depending on the package ${PN}-libbz2 rather than the library, but I'd like to know why this case would be different than all the others. Or am I completely misunderstanding you?

esben: No, I think you get my point. I just wanted to raise the concern. But with that done, I am fine with your change :+1:

Villemoes: It's a valid point, and one I hadn't thought about. But until we see an example where this breaks, I think I'll prefer following "depend on the item, not the package". We could try to find a way that dependencies would prefer items from the same recipe.

PR #477, opened 2016-12-22
busybox: add version 1.26.0 (unstable)

sknsean:
Need to add priority -1 before merging...


sknsean:
quite interresting to diff when I don't have included the removal of 1.25.1 :-S

recipes/busybox/busybox_1.26.0.oe
```diff
@@ -0,0 +1,6 @@
+require busybox.inc
+
```
esben: Perhaps a `PRIORITY=-1` to avoid this unstable version to become the default.

esben:
1.26.2 is out now...

PR #480, opened 2017-01-05
busybox: add use flag to disable external dhcp client startup

vmbracvlvm:
Set 'busybox_disable_external_dhcp_startup' to False to disable external
dhcp client startup.


recipes/busybox/busybox-configure.inc
```diff
@@ -317,6 +317,14 @@ do_configure_busybox_dhclient () {
 BUSYBOX_SIMPLE_USE_FLAGS += "ifupdown_external_dhcp=1::\
 CONFIG_FEATURE_IFUPDOWN_EXTERNAL_DHCP"
 
+RECIPE_FLAGS += "busybox_disable_external_dhcp_startup"
```
esben: Why the negated USE flag?

Why not `USE_busybox_ifupdown_external_dhcp` and set it to default `True`?

PR #483, opened 2017-01-12
valgrind: re-add arm-cortex-support.patch

Villemoes:
The patch to make valgrind's configure system recognize arm-cortexa* was
omitted from the upgrade to 3.12, but it is still needed. Add it back in
a slightly modified form

- only patch the .ac file since we now do autoreconf anyway
- add a patch header
- put it into non-version specific subdirectory, since we're likely
  going to need this forever
- only apply it when the target cpu is indeed arm

### DIFF
--- a/recipes/valgrind/files/arm-cortex-support.patch
+++ b/recipes/valgrind/files/arm-cortex-support.patch
@@ -1,0 +1,29 @@
+Valgrind's configure script checks for ${host_cpu} starting with
+armv7, but we call it just arm. Since Cortex A chips are (at least)
+ARMv7, make the configure script recognize arm-cortexa*.
+
+Origin: OE-lite (first introduced in 5b66593c)
+Upstream-status: Not attempted
+
+--- valgrind-3.12.0/configure.ac.orig	2017-01-11 08:28:48.965259010 +0000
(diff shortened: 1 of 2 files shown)


recipes/valgrind/files/arm-cortex-support.patch
```diff
@@ -0,0 +1,29 @@
+Valgrind's configure script checks for ${host_cpu} starting with
+armv7, but we call it just arm. Since Cortex A chips are (at least)
+ARMv7, make the configure script recognize arm-cortexa*.
+
+Origin: OE-lite (first introduced in 5b66593c)
+Upstream-status: Inappropriate
```
kimrhh: is this just what you think, or have we checked with upstream? not all upstreams are refusing changes like this :)

if we just don't want to spend time on trying to upstream it, which of course can be reasonable, then we should write something else there

Villemoes: Hm, yocto does something different, which doesn't require a patch but is a little hacky:
```
# valgrind checks host_cpu "armv7*)", so we need to over-ride the autotools.bbclass default --host option
EXTRA_OECONF_append_arm = " --host=armv7${HOST_VENDOR}-${HOST_OS}"
```
Not sure what the best solution really is. But I'll change the Upstream-status appropriately if I stick with the patch.

PR #485, opened 2017-01-12
Add a couple of small features to dhcpcd recipe

kimrhh:


recipes/dhcpcd/files/init
```diff
@@ -16,7 +16,12 @@ do_start() {
...
 
 do_stop() {
-    start-stop-daemon -K --pidfile $PIDFILE --exec $DAEMON
+    # only try to stop the daemon if there is actually something
+    # running. Otherwise a restart before a start will fail
+    start-stop-daemon -K -o --pidfile $PIDFILE --exec $DAEMON
+    if [ "$?" != "0" ] ; then
+	start-stop-daemon -K --pidfile $PIDFILE --exec $DAEMON
+    fi
 }
```
Villemoes: Does this work? You still have `set -e`, so if the first start-stop-daemon fails, you should never get around to testing $?. At least that how a quick test in bash behaves. If you make the do_stop part of a conditional chain, the set -e behaviour is suppressed, so it seems (again, tested in bash) that one can do "do_stop || true" in the restart) case. But then you don't need to change do_stop itself.

recipes/dhcpcd/dhcpcd.inc
```diff
@@ -68,3 +68,14 @@ do_install_initscript() {
...
+RECIPE_FLAGS += "dhcpcd_args"
+DEFAULT_USE_dhcpcd_args = " "
+do_install_config() {
+    install -d ${D}/${sysconfdir}/default
+    echo 'ARGS="${USE_dhcpcd_args}"' > ${D}/${sysconfdir}/default/dhcpcd
+}
+do_install[prefuncs] += "do_install_config"
+PACKAGES =+ "${PN}-conf"
+FILES_${PN}-conf = "${sysconfdir}/default/dhcpcd ${sysconfdir}/dhcpcd.conf"
+RDEPENDS_${PN} += "${PN}-conf"
```
Villemoes: nit: missing newline at end of file.

recipes/dhcpcd/files/init
kimrhh: it does work, since the -o just "checking if work is to be done" according to the docs, which means we only really stop it if there is something to stop. It is surely not the most obvious way to solve this, but dropping set -e means we need to change it a lot more to handle errors

recipes/dhcpcd/dhcpcd.inc
kimrhh: fixed :)

recipes/dhcpcd/files/init
Villemoes: The docs say "-o,--oknodo             Exit with status 0 if nothing is done", which I read as saying that the first start-stop-daemon does attempt to stop $DAEMON - it just also returns 0 if it doesn't succeed in doing so. And it seems that this means "/etc/init.d/dhcpcd stop" will always return 0, which I don't like (but I'm fine with making restart mean "restart or start"). But _regardless_ of the semantics of -o, the last three lines are dead code, because if the first start-stop-daemon exits with 0, we do nothing, and if it exits non-zero, the script exits right there.

kimrhh: you are right the three lines are dead code, but I don't think you are right that stop will always return 0 when we add the -o , there is surely a difference between "no work done" and "there was an error when stopping", but will verify tomorrow to be sure.

I don't think it is a great idea to throw away an error when stopping with || true

Villemoes: Yes, it would be absolutely ideal to have the action "stop" exit non-zero if an error occured or no daemon was running, while "restart" would ignore the latter case and proceed to the start part. However, the source doesn't suggest that's how it works:
```
	if (opt & CTX_STOP) {
		int i = do_stop();
		return (opt & OPT_OKNODO) ? 0 : (i <= 0);
	}
```

PR #486, opened 2017-01-12
util-linux: ncurses fixups

Villemoes:
When trying to compile with USE_ncurses set, I hit lots of weird
errors. I think part of the problem is that --with-ncursesw is 'auto' by
default, so it is not enough to disable that in the !USE_ncurses case,
it very much also has to be explicitly disabled when we want to use the
--with-ncurses option. Otherwise, the configure script tries
to run ncursesw-config, which it happened to pick up from the host, and
then it obviously fails when it can't link with libncursesw.

But we also have to explicitly disable widechar support to convince the
configure script that --with-ncurses is ok, otherwise one gets a
non-helpful

  wide-char support enabled, but non-wide ncurses selects

For me (with MACHINE_CPU = "arm-926ejs"), the end result passes
packageqa with and without USE_ncurses set.

### DIFF
--- a/recipes/util-linux/util-linux.inc
+++ b/recipes/util-linux/util-linux.inc
@@ -6,7 +6,7 @@ RECIPE_TYPES = "machine native sdk"
 
 COMPATIBLE_HOST_ARCHS = ".*linux"
 
-inherit autotools autotools-autoreconf gettext
+inherit autotools autotools-autoreconf gettext pkgconfig
 
 require conf/fetch/kernelorg.conf
 MA = "${@d.get('PV').split('.')[0]}"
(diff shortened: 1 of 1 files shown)


kimrhh:
what is ncursesw supposed to be used for? just wondering if we break something by hard-disabling it always

Villemoes:
It's for choosing whether to build against libncurses or libncursesw, but AFAICT we've never built the latter, so I don't think it can break. Note that this does nothing in the !USE_ncurses case (because that already had both --without settings).

kimrhh:
if it really was automagic before, you can't really be sure we did not build libncursesw, but it should be fixed anyway of course. just wondering if we could figure out what might have picked it up if it was built, but I have NFC about ncurses usage :)

kimrhh:
seems it should only be used if e.g. someone tried to enable utf-8 support for a full oe-lite project. To my knowledge nobody ever tried that. So more configuration can be added if someone will work on that some day

Villemoes:
It wouldn't be built automagically just because util-linux's configure script would want it (because it happened to pick up a ncursesw5-config from the host). It would be both awesome and terrible if DEPENDS were set based on an initial configure run. The "more configuration can be added" is probably what is meant by the fixme (which I didn't introduce).

kimrhh:
it has absolutely nothing to do with DEPENDS. If autotools picks up whatever is needed from the host and auto-enables functionality, then it is built. Lots of examples of this. I have not checked if this is the case here but you initial message suggest it is.

moths:
This requires oe-lite/core#210 ncurses pkgconfig/binconfig fixups. If not, you get "cannot find -ltinfo"

esben:
oe-lite/core#210 have been merged, dso this should be ready to go in.

PR #491, opened 2017-01-16
Revert "gdb: remove version 7.10.1"

Villemoes:
This reverts commit cc3ad55bbb6e3161b54880aab38924536cb6c8e8, and sets
the priority for gdb 7.11 to -1 to make OE-lite pick up 7.10.1 by
default (unless one explicitly depends on 7.11).

With 7.11, our internal fulltest fails with

   thread.c:982: internal-error: is_thread_state: Assertion `tp' failed.
   A problem internal to GDB has been detected,
   further debugging may prove unreliable.

I don't see any of the patches we apply on top of 7.11 touching that
file, so I don't know if one of those are to blame, or if it's simply
7.11 which is broken - google only turns up with ancient hits, much
older than 7.11. Whatever the cause, reverting to 7.10.1 fixes the
issue.

Signed-off-by: Rasmus Villemoes <rasmus.villemoes@prevas.dk>

### DIFF
--- a/recipes/gdb/gdb_7.10.1.oe
+++ b/recipes/gdb/gdb_7.10.1.oe
@@ -1,0 +1,1 @@
+require gdb.inc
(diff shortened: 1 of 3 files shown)


esben:
Can you retest with 7.11.1 and 7.12?

PR #494, opened 2017-01-22
Avahi bluez boost

moths:


kimrhh:
seems there is a build issue for the arm archs, are you sure about that BJAM addition? :)

moths:
enabling boost module "context" in oe-lite:
`libs/context/src/unsupported.cpp:7:2: error: #error "platform not supported"`

In openembedded:
`# only supported by x86 and powerpc`
`BOOST_LIBS_append_x86 = " context coroutine"`

Works in oe-lite for 1.59 with 32 bit and 64 bit ARM (tested on 32 bit target only):
`BJAM_OPTIONS += "  abi=aapcs "`

Dropping 1.63, which have other issues, and adding abi=aapcs to aarch64.

kimrhh:
:+1:

recipes/avahi/avahi_0.6.31.oe
```diff
@@ -1,8 +0,0 @@
-require avahi.inc
-SRC_URI += "\
-file://reuseport-check.patch \
-file://0001-avahi-fix-avahi-status-command-error-prompt.patch \
-file://out-of-tree.patch \
-file://avahi_fix_install_issue.patch \
-file://fix_for_automake_1.12.x.patch \
-"
```
Villemoes: Are all the issues these patches solved fixed in 0.6.32?

moths: out-of-tree.patch: fixed in 0.6.32
avahi_fix_install_issue.patch: fixed in 0.6.32
fix_for_automake_1.12.x.patch: fixed in 0.6.32

removing them from files/ dir.

PR #497, opened 2017-01-23
busybox: Add s6-rc integration of hwclock

esben:
This adds an s6-rc service called rtc, which reads hardware clock (RTC)
on start, and writes it on stop.  It can be handy as startup and shutdown
service.

If /etc/hwclock/UTC file exists and does contain something else than the
string "1", the time hwclock is assumed to contain local time. Otherwise,
it is assumed to contain UTC time.

If /etc/hwclock/DEVICE file exists, it specifies the RTC device to use.
Otherwise, /dev/rtc is used.

### DIFF
--- a/recipes/busybox/busybox.inc
+++ b/recipes/busybox/busybox.inc
@@ -75,6 +75,11 @@ do_compile_binfmt_flat_fixup() {
 
 require busybox-install.inc
 
+inherit s6rc
+SRC_URI:>USE_s6rc = " file://rtc.up file://rtc.down"
+S6RC_ONESHOT_SERVICES:>USE_busybox_hwclock = " rtc"
+RECIPE_FLAGS += "rtc_s6rc_dependencies"
+
 PACKAGES:<USE_busybox_udhcpd = "${PN}-udhcpd-config "
 PROVIDES_${PN}-udhcpd-config = " udhcpd-config"
 FILES_${PN}-udhcpd-config = "${sysconfdir}/udhcpd.conf "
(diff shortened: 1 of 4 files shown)


recipes/busybox/busybox.inc
```diff
@@ -75,6 +75,11 @@ do_compile_binfmt_flat_fixup() {
 
 require busybox-install.inc
 
+inherit s6rc
+SRC_URI:>USE_s6rc = " file://rtc.up file://rtc.down"
+S6RC_ONESHOT_SERVICES:>USE_busybox_hwclock = " rtc"
+RECIPE_FLAGS += "rtc_s6rc_dependencies"
```
kimrhh: where is this flag used?

recipes/busybox/busybox_1.25.1.oe
```diff
@@ -3,5 +3,4 @@ require busybox.inc
 SRC_URI += "file://dhclient-options.patch"
 SRC_URI += "file://mdev-create-devices-from-sys-dev.patch"
 
-RECIPE_FLAGS += "s6rc"
```
kimrhh: should this really be removed with the rest of the s6 related logic still there?

recipes/busybox/busybox.inc
esben: This is used by s6rc.oeclass, allowing for (fx. DISTRO) configuration of service dependencies for rtc service.

recipes/busybox/busybox_1.25.1.oe
esben: Yes.  It is now added as part of inherit of s6rc.oeclass.

esben:
@kimrhh Comments addressed...

PR #503, opened 2017-01-31
Glibmm, official C++ interface for Glib

moths:


recipes/glibmm/glibmm.inc
```diff
@@ -0,0 +1,33 @@
...
+
+DEPENDS += "\
+native:perl native:util/mm-common-prepare \
+mm-common-dev \
+libstdc++ libsigc++ \
+libglib-2.0 libgobject-2.0 libgmodule-2.0 \
+libgthread-2.0 	libgio-2.0 \
+"
+
+do_install[postfuncs] += "do_install_remove_unused"
```
Villemoes: Where is that function defined?

moths: Yes, left over from workaround before core merged oe-lite/core#216, removed now.

PR #504, opened 2017-02-01
openvpn 2.4.0

vmbracvlvm:


recipes/openvpn/openvpn_2.4.0.oe
```diff
@@ -0,0 +1,6 @@
+require ${PN}.inc
+
+EXTRA_OECONF += "--disable-plugin-auth-pam"
+
+EXTRA_OECONF += "IPROUTE=${base_sbindir}/ip --enable-iproute2"
```
esben: Do we really need to specify IPROUTE?

vmbracvlvm: In my experience yes. If not set configure will try to do it.

esben: Ouch, so ./configure will look it up on host, and build in that path?

PrevasJsch:
Actually it seems that specifying IPROUTE is only needed when building in my customized manifest. When building against oe-lite master branches it seems to work.

PR #510, opened 2017-02-17
ntp fixup

esben:


recipes/ntp/ntp.inc
```diff
@@ -77,7 +77,7 @@ DEFAULT_USE_ntpd_sysvinit_stop	= "25"
 
 inherit auto-package-utils
 AUTO_PACKAGE_UTILS = "ntpd ntpdate ntp-wait ntp-keygen sntp ntpq ntpdc \
-    ntptime ntptrace tickadj"
+    ntptime ntpsweep ntptrace tickadj calc_tickadj update-leap"
```
Villemoes: Just curious: I see the ntpsweep perl script in the build/ dir, but it doesn't seem to get installed. Does it need some special configure option, or does it magically detect that the target doesn't have perl, or?

recipes/ntp/ntp.inc
```diff
@@ -30,6 +30,8 @@ EXTRA_OECONF_CRYPTO = "--without-crypto"
 EXTRA_OECONF_CRYPTO:USE_crypto = "--with-crypto"
 DEPENDS:>USE_crypto = " libcrypto"
 
+EXTRA_OECONF += "--disable-debugging"
```
Villemoes: OK, but since I currently happen to be doing ntp debugging on a project, can I selfishly ask you to add a USE flag to enable it?

recipes/ntp/ntp.inc
esben: It gets installed to usr/bin here.  Magic...

recipes/ntp/ntp.inc
esben: Done.

esben:
No, this should affect use of ntpd to steer the local clock using remote peers.

recipes/ntp/ntp.inc
esben: Looking through the autoconf magic, it looks like it should get installed by default, but can be disabled by setting NTPSWEEP_NI environment variable when calling ./configure

PR #512, opened 2017-03-17
busybox-mdev: use existing devtmpfs for /dev if mounted

vmbracvlvm:
If devtmpfs is already mounted on /dev use this instead of creating and
repopulating new tmpfs.

### DIFF
--- a/recipes/busybox/files/busybox-mdev
+++ b/recipes/busybox/files/busybox-mdev
@@ -36,8 +36,12 @@ fi
 
 #mdev_dash_s SECTION_START
 # Create tmpfs for /dev
-echo "Creating tmpfs at /dev"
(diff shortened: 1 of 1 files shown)


recipes/busybox/files/busybox-mdev
```diff
@@ -36,8 +36,12 @@ fi
 
 #mdev_dash_s SECTION_START
 # Create tmpfs for /dev
-echo "Creating tmpfs at /dev"
-mount -t tmpfs tmpfs /dev -o size=64k,mode=0755
+if mount | grep -q 'devtmpfs on /dev type devtmpfs'; then
```
esben: You could also use `mountpoint` command, but we unfortunately have it disabled in default busybox config :disappointed:

PR #515, opened 2017-03-24
pcre: Fixup library packaging

esben:
The libc RDEPENDS is managed directly in the DEPENDS_${PN}-lib*
assignments, the removed line was unneeded.

The AUTO_PACKAGE_LIBS_DEV_RDEPENDS assignment ensures that the pcre-dev
package gets installed when fx. target:libpcre-dev is added to RDEPENDS
in an sdk-image recipe.

The FILES_* assignments places the header files in the corresponding
library specific -dev package.

### DIFF
--- a/recipes/pcre/pcre.inc
+++ b/recipes/pcre/pcre.inc
@@ -49,9 +49,12 @@ DEPENDS_${PN}-pcretest += "libc libpcre libpcreposix"
 
 inherit auto-package-libs
 AUTO_PACKAGE_LIBS = "pcrecpp pcreposix pcre"
-AUTO_PACKAGE_LIBS_RDEPENDS += "libc"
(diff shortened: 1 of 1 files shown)


esben:
Sort of.  From aut-package-libs.oeclass:

```python
        pkg_depends = (d.get("DEPENDS_" + pkg) or "").split()

        pkg_rdepends = d.get("RDEPENDS_" + pkg, True)
        if pkg_rdepends is None:
            pkg_rdepends = []
            for dep in pkg_depends:
                if dep.startswith("lib"):
                    pkg_rdepends.append(dep)
```

Villemoes:
Ew. I see why that may be convenient, but it also means that setting one RDEPENDS disables that auto-smartness. The patch is fine, but please expand the first paragraph with some info about this subtlety.

PR #516, opened 2017-03-28
stress-ng: new recipe

Villemoes:
There's a build problem on ARM with binutils < 2.26 (on my host which
has 2.26, I cannot reproduce it, but current meta/core is at 2.25). The
included patches are from upstream, so shouldn't be needed in the next
release.

### DIFF
--- a/recipes/stress-ng/stress-ng.inc
+++ b/recipes/stress-ng/stress-ng.inc
@@ -1,0 +1,53 @@
+SUMMARY = "stress-ng tests a system by stressing the physical and kernel interfaces."
+
+DESCRIPTION = "stress-ng will stress test a computer system in various selectable ways. It was designed to exercise various physical subsystems of a computer as well as the various operating system kernel interfaces."
+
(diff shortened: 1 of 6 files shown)


recipes/stress-ng/stress-ng.inc
```diff
@@ -0,0 +1,53 @@
+SUMMARY = "stress-ng tests a system by stressing the physical and kernel interfaces."
+
+DESCRIPTION = "stress-ng will stress test a computer system in various selectable ways. It was designed to exercise various physical subsystems of a computer as well as the various operating system kernel interfaces."
+
+LICENSE = "GPLv2"
```
esben: Should be "GPL-2.0".

Villemoes: Thanks. I somehow always end up using the GPLv2 since that seems more natural to me, but I'll have to unlearn that. I've updated it (and reading the source files more carefully, it's actually GPL-2.0+).

PR #518, opened 2017-03-31
add chrony recipe

Villemoes:
I haven't had the opportunity to test this on any target yet, but at least it build pretty cleanly.


vmbracvlvm:
Use spdx compatible licence (GPL-2.0) :)

PR #524, opened 2017-04-20
Lighttpd upgrade

esben:


recipes/lighttpd/lighttpd-1.4.45/0001-Use-pkg-config-for-pcre-dependency-instead-of-config.patch
```diff
@@ -0,0 +1,37 @@
+From e7a8c925b9316a72bdc8f32789ffe56fda5c4788 Mon Sep 17 00:00:00 2001
+From: Alexander Kanavin <alex.kanavin@gmail.com>
+Date: Fri, 26 Aug 2016 18:20:32 +0300
+Subject: [PATCH] Use pkg-config for pcre dependency instead of -config script.
+
+RP 2014/5/22
+Upstream-Status: Pending
```
Villemoes: I find that kinda hard to believe, given that the patch has apparently (judging by the RP line) been in circulation for three years [goes checking] yep, it was added to yocto back when they had a recipe for 1.4.35.

esben: Pending means "No determination has been made yet or not yet submitted to upstream".
See http://openembedded.org/wiki/Commit_Patch_Message_Guidelines which I think is what
we have adopted.

Villemoes: I see, I didn't know they actually had defined meanings, and I've always read "Pending" more or less as meaning what "Submitted" means. I guess the latter is a better name anyway. Sorry for the noise, then.

PR #529, opened 2017-05-04
openssl: Fix parallel make bug in do_compile

esben:

### DIFF
--- a/recipes/openssl/openssl-1.0.2/openssl-fix-race-condition-when-symlink-shared-libs.patch
+++ b/recipes/openssl/openssl-1.0.2/openssl-fix-race-condition-when-symlink-shared-libs.patch
@@ -1,0 +1,45 @@
+From bcea66251e8014c12db81ff54f419fe82b6f8527 Mon Sep 17 00:00:00 2001
+From: Ryan Barnett <ryan.barnett@rockwellcollins.com>
+Date: Thu, 19 Nov 2015 10:47:00 -0600
+Subject: [PATCH] makefile.shared: fix race condition when symlinking shared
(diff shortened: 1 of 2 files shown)


Villemoes:
This seems like the wrong fix. Assuming `b` doesn't exist to begin with, I do think 2 parallel `ln -sf a b` are both guaranteed to succeed (though I'm not entirely convinced), but if `b` may exist (alternatively, 3 or more instances in play), both of these can easily happen:
```
                 symlink(a, b)
symlink(a, b)
unlink(b)
                 unlink(b) # ENOENT
symlink(a, b)
```
```
                 symlink(a, b)
                 unlink(b)
symlink(a, b)
unlink(b)
symlink(a, b)
                 symlink(a, b) # EEXIST
```
Running two instances of `while ln -sf a b ; do : ; done` in two terminals show both of these surprisingly quickly (considering the small size of the race window compared to the time needed to exec an entire process).

I haven't looked closer, but maybe something like https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/openssl/files/openssl-1.0.2a-parallel-symlinking.patch would (also) work?

esben:
Why not just remove the `rm -f` part, and change the `ln -s` to `ln -sf`.  Or is `ln -sf` not atomic?

Villemoes:
No, ln -sf is not atomic. ln does exactly the sequence I wrote above (it almost has to, per posix): (1) symlink, if ok quit, on any error except EEXIST and -f bail, otherwise: (2) unlink; on any error bail out, otherwise: (3) symlink, on any error bail out.

PR #531, opened 2017-05-15
ntp: add capability and s6-init support

Villemoes:


recipes/ntp/ntp.inc
```diff
@@ -98,3 +106,10 @@ FILES_${PN}-ntpdate = "${sysconfdir}/network/if-up.d/ntpdate"
 # ntp originally includes tickadj. It's split off for inclusion in
 # small firmware images on platforms with wonky clocks (e.g. OpenSlug)
 RDEPENDS_${PN}-ntpd = "${PN}-tickadj ${PN}-config ${PN}-init"
```
Villemoes: I can add a patch splitting that off.

recipes/ntp/ntp.inc
```diff
@@ -98,3 +106,10 @@ FILES_${PN}-ntpdate = "${sysconfdir}/network/if-up.d/ntpdate"
...
+
+inherit s6rc
+S6RC_LONGRUN_SERVICES = "ntpd ntpd-log"
+SRC_URI += "file://ntpd.run file://ntpd-log.run"
+RECIPE_FLAGS += "ntpd_s6rc_dependencies"
+DEFAULT_USE_ntpd_s6rc_dependencies = "ntpd-log net"
```
Villemoes: Hm, not sure about that one. It's up to the distro to make sure there's something that provides a "net" service, ntpd shouldn't force that to be the netbase recipe. Alternatively the distro/BSP may set ntpd_s6rc_dependencies to something that makes sense for that distro/BSP.

esben: Fair enough.

recipes/ntp/ntp.inc
esben: That would be nice.  Will you add it to this PR, or make a new one?

Villemoes: Just added it, it does kinda belong with the one adding support for an alternative init.

PR #533, opened 2017-05-15
openssl: fix symlink creation race

Villemoes:
From
<https://gitweb.gentoo.org/repo/gentoo.git/plain/dev-libs/openssl/files/openssl-1.0.2a-parallel-symlinking.patch>.

A slightly cleaner (IMO) alternative to #529.

### DIFF
--- a/recipes/openssl/openssl-1.0.2/parallel-symlinking.patch
+++ b/recipes/openssl/openssl-1.0.2/parallel-symlinking.patch
@@ -1,0 +1,63 @@
+https://rt.openssl.org/Ticket/Display.html?id=3780&user=guest&pass=guest
+
+From cc81af135bda47eaa6956a0329cbbc55bf993ac1 Mon Sep 17 00:00:00 2001
+From: Mike Frysinger <vapier@gentoo.org>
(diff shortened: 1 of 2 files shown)


esben:
FWIW, playing around with doing multiple `while true ; do ln -sf foo bar; done` in parallel, causes problems when used with busybox ln, but not with the coreutils ln I have on my Arch Linux.

Villemoes:
That's probably
```
commit 376967889ed7ed561e46ff6d88a66779db62737a
Author: Paul Eggert <eggert@cs.ucla.edu>
Date:   Sat Feb 11 23:12:31 2017 -0800

    ln: replace destination links more atomically
```
which implements -f in the only race-free way possible
```
+/* A naive "ln -f A B" unlinks B and then links A to B.  This module
+   instead links A to a randomly-named temporary T in B's directory,
+   and then renames T to B.  This approach has a window with a
+   randomly-named temporary, which is safer for many applications than
+   a window where B does not exist.  */
```
That's probably not POSIX-kosher, but there are obviously far more applications that break due to ln -sf not being atomic than due to a temporary file.

PR #534, opened 2017-05-16
dhcp: add s6 init support

Villemoes:
--no-pid is self-explanatory

-f means don't fork

-d means log to stderr (which actually implies -f, but it's better to be
explicit).

Unfortunately, there's no option to make dhcpd not use syslog(3), so
that would require patching dhcpd. Alternatively, one can decide that
e.g. LOG_LOCAL7 (and making sure to set 'log-facility local7' in
dhcpd.conf) messages should be filtered away by one's syslogd
replacement, so that they at least don't take up disk space twice.

Signed-off-by: Rasmus Villemoes <rasmus.villemoes@prevas.dk>

### DIFF
--- a/recipes/dhcp/dhcp-4.3.5/disable-syslog.patch
+++ b/recipes/dhcp/dhcp-4.3.5/disable-syslog.patch
@@ -1,0 +1,233 @@
+Disable syslog the hard way: Put #if 0 around all calls of syslog and
+other functions declared in syslog.h.
+
+Also #if 0 out #includes of syslog.h itself, so that future additions
+of syslog calls hopefully make the build fail due to missing function
+or macro declarations.
+
+Exempt stables.c from this, to allow existing configuration files with
+a log-facility stanza to still parse.
+
+Origin: OE-lite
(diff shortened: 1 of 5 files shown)


esben:
Hmmm...  How ugly would such a patch (avoiding syslog spamming) look like?

Villemoes:
Not sure. Alternatively one can LD_PRELOAD a fakesyslog.so that essentially turns vsyslog() into vprintf(stderr), plus some bells&whistles to handle the case where the daemon unconditionally closes the stdio descriptors (e.g. let the supervisor pass a FAKESYSLOG_FD via the environment indicating where syslog output should go).

esben:
Do we have fakesyslog.so?

Villemoes:
No, we don't - it was just a thought, and something that might be useful for other daemons as well. FWIW, I started doing a "add `#if 0` around all syslog/openlog/closelog/setlogmask calls" patch, and it compiles fine. But dhcp links swith libisc from the bind recipe, which also introduces syslog calls... I'll post the patch I've done with a note that it doesn't quite eliminate syslog spamming.

PR #535, opened 2017-05-17
some wtmp handling

Villemoes:
Add a wtmptrim util, use it in s6-init's init-utmp, and create /var/log/btmp with appropriate permissions.

### DIFF
--- a/recipes/misc/wtmptrim.oe
+++ b/recipes/misc/wtmptrim.oe
@@ -1,0 +1,20 @@
+SUMMARY = "Utility for keeping the tail of wtmp/btmp files"
+LICENSE = "GPL-2.0+"
+SRC_URI = "file://wtmptrim.c"
+COMPATIBLE_HOST_ARCHS = ".*linux"
+
+inherit c
+
+CFLAGS += "-Wall -Wextra -Werror -g"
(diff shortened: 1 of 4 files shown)


recipes/misc/wtmptrim.oe
```diff
@@ -0,0 +1,19 @@
+SUMMARY = "Utility for keeping the tail of wtmp/btmp files"
+LICENSE = "GPL-2.0+"
+SRC_URI = "file://wtmptrim.c"
+COMPATIBLE_HOST_ARCHS = ".*linux"
+
+inherit c
+
+S = "${SRCDIR}"
+do_compile() {
+    ${CC} -Wall -Wextra -Werror -O2 -g -o wtmptrim wtmptrim.c
```
esben: I think you should use `${CFLAGS}` here.

Villemoes: As in, set CFLAGS to `-Wall -Wextra -Werror -O2 -g` and substitute that, or append `Wall -Wextra -Werror` to the existing CFLAGS, or just use the existing CFLAGS without adding `-W*`?

esben: Append (which you have done already now).

PR #536, opened 2017-05-18
vsftpd: add s6 init support

Villemoes:
vsftpd has no way of getting it to log to stdout/stderr. Instead, we
have to tell it to write to the fifo s6-log is reading from. Setting
these on the command line means one doesn't have to deal with this in
the .conf file. It doesn't actually appear to ever write anything to
stdout/stderr, so one might want to redirect those to /dev/null, but I
suppose having them point at the fifo doesn't hurt.

Note that vsftpd reads the default configuration file (/etc/vsftpd.conf)
after parsing the command line if (and only if) no configuration files
were given on the command line.

The one logging-related thing one might do in the .conf file is to set
xferlog_std_format=YES if one wants wu-ftpd compatible logs (whatever
that means).

### DIFF
--- a/recipes/vsftpd/files/vsftpd.run
+++ b/recipes/vsftpd/files/vsftpd.run
@@ -1,0 +1,9 @@
+#!/bin/execlineb -P
+redirfd -w 1 /run/vsftpd-log.fifo
+fdmove -c 2 1
+/usr/sbin/vsftpd
+  -obackground=NO -olisten=YES
+  -ovsftpd_log_file=/run/vsftpd-log.fifo
+  -oxferlog_file=/run/vsftpd-log.fifo
+  -oxferlog_enable=YES
+  -oxferlog_std_format=NO
--- a/recipes/vsftpd/files/vsftpd-log.run
+++ b/recipes/vsftpd/files/vsftpd-log.run
@@ -1,0 +1,9 @@
+#!/bin/execlineb -P
+if { if -t -n { test -e /var/log/vsftpd }
+  install -d -o nobody -g nogroup -m 755 /var/log/vsftpd }
+if { if -t -n { test -e /run/vsftpd-log.fifo }
+  mkfifo /run/vsftpd-log.fifo }
+redirfd -rnb 0 /run/vsftpd-log.fifo
+s6-setuidgid nobody
+exec -c
+s6-log -b -- s131072 t /var/log/vsftpd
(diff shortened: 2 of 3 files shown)


recipes/vsftpd/files/vsftpd.run
```diff
@@ -0,0 +1,9 @@
+#!/bin/execlineb -P
+redirfd -w 1 /run/vsftpd-log.fifo
+fdmove -c 2 1
+/usr/sbin/vsftpd
+  -obackground=NO -olisten=YES
+  -ovsftpd_log_file=/run/vsftpd-log.fifo
+  -oxferlog_file=/run/vsftpd-log.fifo
```
esben: Hmm...  So mixing "normal" log output with transfer log information...  Maybe add a vsftpd-xferlog service to keep them separated.

Villemoes: No, only one of these is actually used. It's confusing, so let me quote all the relevant config options:

```
       xferlog_enable
              If enabled, a log file will be maintained detailling uploads and downloads.  By default,
              this file will be placed at /var/log/vsftpd.log, but this  location  may  be  overridden
              using the configuration setting vsftpd_log_file.

              Default: NO (but the sample config file enables it)

       xferlog_std_format
              If enabled, the transfer log file will be written in standard xferlog format, as used by
              wu-ftpd. This is useful because you can reuse existing transfer  statistics  generators.
              The default format is more readable, however. The default location for this style of log
              file is /var/log/xferlog, but you may change it with the setting xferlog_file.

              Default: NO

       vsftpd_log_file
              This option is the name of the file to which we write the vsftpd style  log  file.  This
              log  is  only written if the option xferlog_enable is set, and xferlog_std_format is NOT
              set. Alternatively, it is written if you have set the option dual_log_enable.  One  fur‐
              ther  complication  -  if  you have set syslog_enable, then this file is not written and
              output is sent to the system log instead.

              Default: /var/log/vsftpd.log

       xferlog_file
              This option is the name of the file to which we write the wu-ftpd  style  transfer  log.
              The  transfer  log is only written if the option xferlog_enable is set, along with xfer‐
              log_std_format.   Alternatively,  it  is  written   if   you   have   set   the   option
              dual_log_enable.

              Default: /var/log/xferlog
```
So, except if one set dual_log, only one of these is actually used. They are both used for "transfer log" information.

esben: Ahh :+1:

PR #538, opened 2017-05-18
jq: add use flag for setting --disable-maintainer-mode

Villemoes:
When the host's bison is < 3.0, do_configure complains

  configure: error: You need bison version 3.0 or greater, or use --disable-maintainer-mode.

Add a knob to add that configuration option.

### DIFF
--- a/recipes/jq/jq.inc
+++ b/recipes/jq/jq.inc
@@ -21,3 +21,5 @@ AUTO_PACKAGE_LIBS = "jq"
 AUTO_PACKAGE_LIBS_DEPENDS = "${DEPS}"
 AUTO_PACKAGE_LIBS_RDEPENDS = "${DEPS}"
 FILES_jq-libjq-dev += "${includedir}"
+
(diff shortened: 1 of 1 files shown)


recipes/jq/jq.inc
```diff
@@ -21,3 +21,6 @@ AUTO_PACKAGE_LIBS = "jq"
 AUTO_PACKAGE_LIBS_DEPENDS = "${DEPS}"
 AUTO_PACKAGE_LIBS_RDEPENDS = "${DEPS}"
 FILES_jq-libjq-dev += "${includedir}"
+
+RECIPE_FLAGS += "jq_disable_maintainer_mode"
+EXTRA_OECONF:>USE_jq_disable_maintainer_mode = " --disable-maintainer-mode"
```
esben: Why not just add this unconditionally to EXTRA_OECONF?

Villemoes: because I hadn't found the readme that actually explained what this was about. Yes, it makes sense to set unconditionally. Pushed.

PR #539, opened 2017-05-19
overlay-files: allow symlinking to a directory

Villemoes:
If /etc/openvpn already exists as a symlink to
/etc/openvpn.default, doing

  ln -sf /some/where /etc/openvpn

ends up creating the link where inside /etc/openvpn.default, rather than
making /etc/openvpn itself a symlink to /some/where. Add -n to avoid
dereferencing the target.

If /etc/openvpn is itself a directory, this will still end up creating
the symlink inside that directory, but this at least fixes the common
case where the symlink openvpn->openvpn.default has been created by
do_rstage_overlay_files.

### DIFF
--- a/recipes/overlay-files/files/overlay.sh
+++ b/recipes/overlay-files/files/overlay.sh
@@ -18,7 +18,7 @@ for file in $OVERLAY_FILES ; do
     if test -e "$overlay_file" ; then
 	if test ! -h "$link_file" -o \
 	"`readlink "$link_file"`" != "$overlay_file" ; then
-	    ln -sf "$overlay_file" "$link_file"
+	    ln -sfnT "$overlay_file" "$link_file"
 	fi
     fi
 done


recipes/overlay-files/files/overlay.sh
```diff
@@ -18,7 +18,7 @@ for file in $OVERLAY_FILES ; do
     if test -e "$overlay_file" ; then
 	if test ! -h "$link_file" -o \
 	"`readlink "$link_file"`" != "$overlay_file" ; then
-	    ln -sf "$overlay_file" "$link_file"
+	    ln -sfn "$overlay_file" "$link_file"
```
esben: Why not use `-T` instead, which would cause a failure instead of a silent wrong action in case of $link_file being a plain directory?

Villemoes: Well, we need at the very least also -n. But I've also been confused by the wrong help output from busybox ln:

```
# ln -h
...
        -T      2nd arg must be a DIR
```
that should probably be "must not". But we also need -n, because unlike coreutils ln, busybox ln also takes -n into account when deciding whether the second argument is a directory. That is, in
```
touch a
mkdir d
ln -s d dl
ln -sfT a dl
```
the last command fails with busybox, but not coreutils... I'll update to include -T.

PR #540, opened 2017-05-24
ntp tweaks

Villemoes:
This fixes a few bugs in the default ntp.conf as well as the sysvinit script, and adds simple utilities that can be used to query ntpd for how accurate the system clock is.


Villemoes:
Pushed default mingw fixup.

PR #541, opened 2017-05-30
s6-init: Add timeout to avoid s6-rc state machine block

esben:
The s6-rc state machine will be blocked while the `s6-rc change` command is
running.  Without a timeout, we could thus block forever.

See https://www.mail-archive.com/supervision@list.skarnet.org/msg01683.html

And as reboot is handled as an s6-rc state machine transition, we will not
even be able to reboot in those cases.

This change introduces a hard-coded 10 second timeout service start and stop
commands.  If the timeout fires, any changes happening later will be picked up
by s6-rc on next `s6-rc change` call.

### DIFF
--- a/recipes/s6-init/s6-init/service
+++ b/recipes/s6-init/s6-init/service
@@ -10,11 +10,11 @@ EOF
     }
 
 start() {
-    s6-rc -u change $*
+    s6-rc -u -t 10000 change $*
 }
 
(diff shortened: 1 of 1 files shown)


esben:
When timeout is reached, s6-rc exits `2` with the following on stderr

```
s6-rc: fatal: timed out
```

PR #545, opened 2017-06-07
more ntp/rtc fixes

Villemoes:
This fixes a stupid bug in ntp.conf I introduced while trying to fix another, and patches hwclock to allow preserving synchronized time to almost centisecond precision on some non-x86 platforms.

### DIFF
--- a/recipes/util-linux/util-linux-2.29/hwclock-localtime.patch
+++ b/recipes/util-linux/util-linux-2.29/hwclock-localtime.patch
@@ -3,7 +3,7 @@ compatibility with a busybox that has not been compiled with long
 option support.
 
 Origin: OE-lite
-Upstream-status: Pending
+Upstream-status: Merged [4d670b2, https://github.com/karelzak/util-linux/pull/449]
 
 --- util-linux-2.29/sys-utils/hwclock.c.orig	2017-05-31 08:24:25.571236824 +0200
 +++ util-linux-2.29/sys-utils/hwclock.c	2017-05-31 08:30:24.048583055 +0200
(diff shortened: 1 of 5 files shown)


recipes/util-linux/util-linux-2.29/hwclock-localtime.patch
```diff
@@ -3,7 +3,7 @@ compatibility with a busybox that has not been compiled with long
 option support.
 
 Origin: OE-lite
-Upstream-status: Pending
+Upstream-status: Merged [4d670b2, https://github.com/karelzak/util-linux/pull/449]
```
esben: Great :+1:

recipes/util-linux/util-linux-2.29/hwclock-tweak-delay.patch
```diff
@@ -0,0 +1,100 @@
...
+# Write the system time to the RTC, read it back, compare to NTP time.
+hwclock --utc --systohc --noadjfile
+hwclock --utc --hctosys --noadjfile
+ntpdate -q 0.pool.ntp.org 1.pool.ntp.org
+# Repeat these three steps a few times. If the discrepancy increases
+# consistently with ~0.5s, try to start over but pass --delay 0.0 to
+# the --systohc command and see if the situation improves.
+
+Origin: OE-lite
+Upstream-status: Inappropriate (in this form - significant changes have happened since 2.29).
```
esben: How much work would it take to make a patch against upstream master?

esben: We probably do not want to get stuck here on 2.29 anyway (with 2.30 being current now).

Villemoes: Well, I actually started patching upstream, which is how I found out that that patch wouldn't apply to 2.29, so I redid it against 2.29. The biggest problem is probably documentation, option naming and explaining what this is about (which I mostly left out of this one because the commit messages explain it). Also, it would actually be nice if hwclock itself figured out what an appropriate --delay value is, but that's quite hard... I'll try to send something upstream and see what they say.

PR #547, opened 2017-06-12
openvpn: create bundle consisting of all openvpn@* services

Villemoes:
In addition to creating the openvpn@X services for each X.conf, also
create a bundle simply called openvpn that contains all of these
services.

The lack of -p in the mkdir command is deliberate; this ensures that
we're actually creating a new bundle and do not clobber some other
hook's output (the caller of the hook script is supposed to have created
the ${1} directory). We still need to create an empty contents file
explicitly in case there are no .conf files.

### DIFF
--- a/recipes/openvpn/files/openvpn.hook
+++ b/recipes/openvpn/files/openvpn.hook
@@ -5,9 +5,14 @@ backtick -n hooksdir { dirname $0 }
 import -u hooksdir
 define basename openvpn@
 define templatedir ${hooksdir}/${basename}.d
+define allbundledir ${1}/openvpn
+if { mkdir ${allbundledir} }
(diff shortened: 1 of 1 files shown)


recipes/openvpn/files/openvpn.hook
```diff
@@ -5,9 +5,14 @@ backtick -n hooksdir { dirname $0 }
 import -u hooksdir
 define basename openvpn@
 define templatedir ${hooksdir}/${basename}.d
+define allbundledir ${1}/openvpn
+if { mkdir ${allbundledir} }
+if { redirfd -w 1 ${allbundledir}/type s6-echo "bundle" }
+if { redirfd -w 1 ${allbundledir}/contents exit 0 }
```
esben: nice touch :)

PR #551, opened 2017-08-08
s6-rc sanity checking

Villemoes:
These three patches allow one to do a build-time check for cyclic dependencies and other errors in the s6 services by inheriting s6rc-sanity from one's rootfs recipe.


Villemoes:
Indeed. At first I had an S6RC_SANITY_something variable where one could list pseudo-services, but then it occured to me that, in all the current use cases, that list can be auto-generated. If the need arises, we can always add variables to help guide the logic - but it's better to do that when we see an actual use case.

PR #553, opened 2017-08-09
s6-init: prevent the syslogd filter script from buffering log lines

Villemoes:
The current sed script ends up buffering its output, so that lines
arrive at the s6-log instance in batches, which means that the time
stamps they get have little to do with when the lines were produced, and
moreover there's obviously a greater chance of losing messages that way.

Prepending stdbuf -oL turns out to only partially solve the
problem. While busybox sed does happen to use libc's stdio, a quirk of
its implementation means that it reads line n+1 before processing (and
hence printing) line n, so there would be a delay of one line - and all
lines would be timestamped with a stamp that really belongs to the
following line.

This awk script seems to be equivalent, works with both busybox and
GNU awk, and explicitly flushes stdout after every print.


esben:
Thanks :+1:

PR #558, opened 2017-08-17
skalibs: Fix default path

esben:
This aligns default path with default (root) path in busybox, which is

    /sbin:/usr/sbin:/bin:/usr/bin

### DIFF
--- a/recipes/s6-init/s6-init/init.stage1
+++ b/recipes/s6-init/s6-init/init.stage1
@@ -1,7 +1,7 @@
 #!/bin/execlineb
 
 # Make sure path and cwd is right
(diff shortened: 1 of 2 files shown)


Villemoes:
Doesn't seem to be enough. The PATH setting in init.stage1 is hard-coded as /usr/bin:/bin:/usr/sbin:/sbin, and that is the one that ends up being inherited down to sshd and hence ssh sessions. I don't know what the --with-default-path flag to skalibs actually affects.

Also, you quote the busybox path correctly, but the second item in your new path is not ${sbindir}.

Villemoes:
New version looks ok. I'd still like to know what the skalibs config option actually does, but it really shouldn't hurt to have the whole system agree on what's in $PATH by default.

PR #566, opened 2017-09-26
update skaware

Villemoes:
These patches add the latest versions of the various skaware software. Doing this exercise, I encountered several bugs, some of which are now fixed upstream. But I've seen s6-rc-init crash rather late (causing the board to fail to boot), both with the old version 0.0.2.1 and with the first-attempted new version 0.2.1.2, crashes which I haven't been able track down or reproduce consistently, and I don't have any good reason to believe the problem is fixed. Hence these patches also add some enhancements that should make it easier to debug, should the problem occur again.


recipes/s6-init/s6-init/init.stage1
```diff
@@ -41,7 +41,19 @@ background
   fdmove -c 2 1
 
   # Exec into stage2 script
-  /etc/rc.init $SERVICE
+  #/etc/rc.init $SERVICE
+  foreground {
+    /etc/rc.init $SERVICE
+  }
+  importas status ?
+  if -t { test $status -ne 0 }
```
esben: Why not simply `if -t -n { /etc/rc.init $SERVICE }` ?

recipes/s6-init/s6-init/init.stage1
```diff
@@ -41,7 +41,19 @@ background
   fdmove -c 2 1
 
   # Exec into stage2 script
-  /etc/rc.init $SERVICE
+  #/etc/rc.init $SERVICE
```
esben: Just remove the line. I think we should keep this commit in the PR. System that care about not hanging in the shell should hopefully have a watchdog design that reboots the system.

recipes/skaware/execline.inc
```diff
@@ -39,3 +40,5 @@ LIBRARY_VERSION = "2"
 DEPENDS_${PN}-libexecline += "libc libskarnet"
 RDEPENDS_${PN}-libexecline += "libc libskarnet"
 FILES_${PN}-libexecline-dev += "${includedir}"
+
+CFLAGS += "-g -fstack-protector"
```
esben: Do you have an idea for why Laurent is using `-fno-stack-protector`?

recipes/s6-init/s6-init/init.stage1
Villemoes: OK, I was a little worried that this would/could be a security issue. But it really is useful for debugging.

recipes/s6-init/s6-init/init.stage1
Villemoes: Well, partly because I don't speak execline fluently, partly because actually having the exit status printed has some value (e.g., in my case, I knew something had died from SIGABRT because this printed 134).

recipes/skaware/execline.inc
Villemoes: No, but I suppose his code might be penalized quite heavily. Maybe he can be convinced to make it a configure option instead of hard-coding it to no.

recipes/s6-init/s6-init/init.stage1
esben: Hmm... I guess it could become a security issue. If there is possibility for non-root users to write content which triggers this failure.  Probably best to leave it out.

recipes/skaware/execline.inc
esben: I think it would be best to take a discussion with him on this before throwing the switch.  If he has good reasons for it, we should not undo it without knowing his reasons and taken that into consideration.

recipes/s6-init/s6-init/init.stage1
esben: It would be nice to have it as a USE flag (which developers can then leave on in local.conf).

recipes/s6-init/s6-init/init.stage1
esben: Makes sense.

recipes/s6-init/s6-init/init.stage1
Villemoes: Yeah, maybe that use flag should just be a string naming something to exec into if /etc/rc.init fails, which can then default to /bin/false. Dunno, I'll try something when I update the PR after Laurent makes some bugfix releases, then we'll see.

recipes/skaware/execline.inc
Villemoes: Yes, I was mostly doing this to get debug symbols for valgrind, but thought I'd try throwing in stack-protector as well just to see if that might fire something; but then I found that it's a bit random whether that ends up taking effect (because in some of the software, he adds -fno-stackprotector to CPPFLAGS which precedes CFLAGS in the compile command, in others, it is appended to CFLAGS itself, and there's no other convenient variable to append to that is guaranteed to have precedence).

recipes/s6-init/s6-init/init.stage1
Villemoes: For now I went with adding exec'ing into /etc/rc.init-fallback, which is essentially just a copy of the system-console service - with a twist, as explained in 0326aed.

esben:
LGTM. How close are you to feel confident about merging this?

ravi-prevas:
Hard to say. I really don't like the unexplained crash late in s6-rc-init, but OTOH, I've also seen that (or at least a very similar one) with the old version, so there's probably no reason to expect it to be new. I've been using the new versions since the bugfix releases came out, and haven't had any problems. So maybe the best thing to do is to merge it, and then give a meta/base bump a spin through our private CI tests.

esben:
Yes, let's do that.

PR #571, opened 2017-12-05
LTTng upgrades

esben:


Villemoes:
OK. I've never used lttng so I don't know how all these components fit together, but I assume you've given the end result a spin.

PR #573, opened 2018-01-03
s6-init: Drop redundant s6-rc dependencies

esben:
With init-coldplug service depending on devd service, which depends on
init-devtmpfs and mount-sys, there is no reason for init-coldplug depends on
that by itself.  It only causes more work if these dependencies needs to be
changed.


ravi-prevas:
But they're not redundant? init-coldplug aka 'mdev -s' very much depends on /sys and /dev being mounted. That we _also_ want a hotplug daemon running before doing the coldplug (to avoid missing some event), and that that daemon also happens to depend on those same filesystems being mounted, seems to be irrelevant. Services should list precisely the stuff they depend on, regardless of what stuff recursive dependencies they currently happen to pull in.

esben:
Ah, I thought that `mdev -s` actually relied on the registered hotplug command (i.e. `mdev`) to do the actual job, and as such did not need anything.  But looking into mdev.c I see that I was completely wrong.

I completely agree with your description on which dependencies should be listed, and had misunderstood the way `mdev -s` worked.

Villemoes:
Well, it turns out you're not entirely wrong: This is indeed apparently how modern udev/sysfs interacts/is supposed to work, and mdevd has just been updated to work that way. See the `mdevd-0.0.1.0 - a mdev-compatible uevent manager` thread at `https://www.mail-archive.com/skaware@list.skarnet.org/`, in particular the posts starting from `https://www.mail-archive.com/skaware@list.skarnet.org/msg01134.html`. That doesn't change that `mdev -s` as currently implemented does directly rely on /sys and /dev, but if and when we switch to mdevd/mdevd-coldplug it's worth keeping this in mind.

PR #576, opened 2018-02-20
class/s6rc-sanity: Fix pipeline dependencies on dynamic services

esben:
In order to allow statically defined services (i.e. in /etc/rc) to have a
producer-for/consumer-for relation to a dynamic service (i.e. one created by a
hook in /etc/rc.hooks/), we create dummy longrun services instead of empty
bundles, and take care to add needed producer-for/consumer-for relations.

Hopefully the hooks will also do the right thing.

### DIFF
--- a/classes/s6rc-sanity.oeclass
+++ b/classes/s6rc-sanity.oeclass
@@ -11,16 +11,34 @@ image_preprocess_s6rc_sanity() {
 	# We assume that each executable file in etc/rc.hooks creates
 	# a service or bundle by the same name. Some of the statically
 	# defined services in etc/rc may depend on some of these, so
-	# create empty bundles which cannot contribute to a dependency
-	# cycle.
+	# create a dummy longrun service to satisfy such services.
 	mkdir -p $tmpd/dynamic
 	if [ -d ./etc/rc.hooks ] ; then
 		for f in $(find ./etc/rc.hooks -maxdepth 1 -type f -executable) ; do
 			b=$(basename $f)
 			d=$tmpd/dynamic/$b
 			mkdir -p $d
-			echo bundle > $d/type
-			touch $d/contents
+			echo longrun > $d/type
+			touch $d/run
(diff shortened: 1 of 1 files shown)


classes/s6rc-sanity.oeclass
```diff
@@ -11,16 +11,36 @@ image_preprocess_s6rc_sanity() {
...
+
+			# Also need to handle cases where a statically defined
+			# service is declared as a producer or consumer for a
+			# dynamically created service
+			for pf in $tmpd/static/*/producer-for ; do
+				test -f $pf || continue
+				read consumer < $pf
+				if [ "$consumer" = "$b" ] ; then
+					producer=$(basename $(dirname $pf))
+					echo "$producer" > $d/consumer-for
```
Villemoes: s6-rc 0.4.0.0 has this in its changelog: 
_s6-rc-compile now accepts multiple lines in a consumer-for file: a longrun can now have multiple producers, which means that pipelines are really funnels._ So please use `>>` to populate consumer-for.

classes/s6rc-sanity.oeclass
```diff
@@ -11,16 +11,36 @@ image_preprocess_s6rc_sanity() {
...
+				fi
+			done
+			for cf in $tmpd/static/*/consumer-for ; do
+				test -f $cf || continue
+				read producer < $cf
+				if [ "$producer" = "$b" ] ; then
```
Villemoes: Similarly, this should probably be changed to testing whether the consumer-for file contains `$d`, so I think you can remove the `read` and replace the conditional with an anchored grep command.

esben:
Like this?

Villemoes:
Yes. One cannot be a producer for multiple services, but I suppose it's nice to keep it symmetric. Also, bash apparently just leaves a `$` at the end of a double-quoted string alone, so it's ok. We probably don't have to worry about service names containing grep metachars.

PR #578, opened 2018-02-22
s6-init: Allow blocking mode to service command

esben:
Adding a '-b' option to service command allows blocking on the s6-rc lock
starting and stopping services.  Without this, service command fails when
called while another (s6-rc) program holds the lock, spewing error messages
like

    s6-rc: fatal: unable to take locks: Resource temporarily unavailable

This is generally the safe approach, so it is kept as the default.

But when using service command while changing s6-rc state (for example during
boot), it can be convenient to wait for the s6-rc lock to be released.

A valid use-case is when an application (that is not running as an s6-rc
service!) wants to stop an s6-rc service.  To avoid failure when started
during boot, the -b option is very helpful.

Note: The -b option must not be used in nested s6-rc invocation.  Anything
running with the the s6-rc lock held (for example an s6-rc service script)
will deadlock forever if using service command with -b option!

### DIFF
--- a/recipes/s6-init/s6-init/service
+++ b/recipes/s6-init/s6-init/service
@@ -2,19 +2,21 @@
 
 usage() {
     cat <<EOF
-usage: $0 start <service>...
-       $0 stop <service>...
-       $0 restart <service>...
-       $0 status [<service>]
+usage: $0 [-b] start <service>...
+       $0 [-b] stop <service>...
+       $0 [-b] restart <service>...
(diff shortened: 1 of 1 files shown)


Villemoes:
The change itself is ok, but I generally hate patches that needs to touch a million lines, adding the same code to each. There's no sane way to review whether you've missed a spot... So here we'd be better of if we had had a helper function for calling s6-rc (with additional args/options) and had used that consistently. Oh well.

PR #579, opened 2018-05-02
Stress-ng 0.09.24

esben:


Villemoes:
Interesting... it's rare that one sees this
```
stress-cpu.c:2495:1: internal compiler error: Segmentation fault
```

Villemoes:
I forced the buildbot to do it again, and it segfaults at the same place. So either the source code is truly horrible, or we're building a broken compiler.

PR #585, opened 2018-08-22
libgpiod: New recipe

esben:


Villemoes:
OK, modulo I'd like the buildbot to give a green light as well.

esben:
Yes, it needs updated Linux headers.  I will send a PR for that.

esben:
And it will of-course only work on Linux, but buildbot should go green now.

PR #586, opened 2018-08-24
Allow USE flag configuration of s6-log arguments

esben:


Villemoes:
From a quick look, ok. Any idea what's causing the build failures?

esben:
Hmm... Seems to be some unrelated fetching issues.  I will look into it.

esben:
Argh, sources that has gotten lost.
I managed to locate all missing pieces, except

http://downloads.sourceforge.net/gsoap2/gSOAP/gsoap_2.8.37.zip

which seems to have gone missing on the Internet as such.
Do you perhaps have a mirror copy of this at Prevas?

http://dev.automforge.net/CanFestival-3/archive/8bfe0ac00cdb.tar.gz

which is a really bad SRC_URI.  It is an auto-generated git snapshot.  Downloading it now/again does not give the expected source signature.
Do you perhaps have a mirror copy of this at Prevas?  If not, I will try and fixup the recipe, or remove it if no proper source exists.

esben:
Argh, sources that has gotten lost.
I managed to locate all missing pieces, except

http://downloads.sourceforge.net/gsoap2/gSOAP/gsoap_2.8.37.zip

which seems to have gone missing on the Internet as such.
Do you perhaps have a mirror copy of this at Prevas?

http://dev.automforge.net/CanFestival-3/archive/8bfe0ac00cdb.tar.gz

which is a really bad SRC_URI.  It is an auto-generated git snapshot.  Downloading it now/again does not give the expected source signature.
Do you perhaps have a mirror copy of this at Prevas?  If not, I will try and fixup the recipe, or remove it if no proper source exists.

esben:
Buildbot is green now.

PR #587, opened 2018-08-28
mobile-broadband-provider-info: Fix source SHA1 signature

esben:
This fixes the sha1sum to match the publically announced sha1sum, which
incidentally matches all the tarballs out there.  It is also the same
signature that other projects uses.

I don't have a copy of the old tarball, so I cannot see what has changed.


esben:
As we have multiple source bugs currently, this PR will probably not build.
If the mobile-broadband-provider-info fetch works, I believe it needs to be merged.

@Villemoes If you can find a copy of the old source, it would be interesting to see what has changed.

Villemoes:
I've put a copy of my canfestival and mobile... tarballs in rasmusvillemoes.dk/filer/oe-lite, and checked that their sha1sums match what I have in my current meta/base .sig files. I do remember noticing and discussing the canfestival thing with kiho a long time ago, maybe on gitter, but we didn't come up with any real solution.

Villemoes:
Hm, I wanted to download the new tarball to compare with mine, but wget say 'Username/Password Authentication Failed.', curl gives me a 301, and trying with 'curl -L' to make it follow that ends up with an ascii text file saying "You need to sign in or sign up before continuing."

PR #602, opened 2019-01-28
gettext: Fix parallel build seen with mingw32 builds

esben:

### DIFF
--- a/recipes/gettext/gettext-0.19.4/intl-add-missing-libintl-h-dependency.patch
+++ b/recipes/gettext/gettext-0.19.4/intl-add-missing-libintl-h-dependency.patch
@@ -1,0 +1,42 @@
+commit cb29e273b0f86b6e54fd6844ebfae2028bd39ac2
+Author: Václav Slavík <vaclav@slavik.io>
+Date:   Sun Jan 4 08:39:29 2015 +0900
+
+    intl: Add missing libintl.h dependency from pluralx.$lo
+    
+    plural(x).c depends on libintl.h if ENABLE_NLS, but the makefile
+    didn't have corresponding dependency for pluralx.c, because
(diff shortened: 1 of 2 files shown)


Villemoes:
ravi@prevas-ravi:/mnt/xfs/devel/gettext$ git show cb29e273b0f86b6e54fd6844ebfae2028bd39ac2
commit cb29e273b0f86b6e54fd6844ebfae2028bd39ac2
Author: Václav Slavík <vaclav@slavik.io>
Date:   Sun Jan 4 08:39:29 2015 +0900

    intl: Add missing libintl.h dependency from pluralx.$lo
    
    plural(x).c depends on libintl.h if ENABLE_NLS, but the makefile
    didn't have corresponding dependency for pluralx.c, because
    d2d04ba forgot to add it.  Fix by using $(PLURAL_OBJECT) for the
    dependency rule too.
    * Makefile.in ($(PLURAL_OBJECT)): Depend on libintl.h.

ravi@prevas-ravi:/mnt/xfs/devel/gettext$ git tag --contains cb29e273b0f86b6e54fd6844ebfae2028bd39ac2
v0.19.5
v0.19.5.1
v0.19.6
v0.19.7
v0.19.8
v0.19.8.1

Please backport instead, or upgrade the gettext recipe to a newer version.

esben:
I have backported the fix.  We already have the newest tagged version, but the old version is kept around because it works for mingw32.  I haven't investigated how much work it would be to get 0.19.8.1 to work for mingw32, but I don't think it should block fixing the current old version.

Villemoes:
urgh, including the changelog hunk makes it harder to backport other patches. But as long as we elide the changelog hunks in all the other patches, all should be well :)

esben:
Better to fix mingw32 build for latest version than adding more backports to old version.

